PR #5, opened 2026-02-10
Add macOS universal binary (arm64 + x64) support

kdroidFilter:
## Summary

- Adds `universalBinary { enabled = true; x64JdkPath = "..." }` DSL under `macOS {}` to build fat macOS apps running natively on both Apple Silicon and Intel
- Uses a subprocess strategy: launches a separate Gradle build under Rosetta with an x64 JDK, then merges both distributables with `lipo`
- Registers `mergeUniversalBinary`, `packageUniversalDmg/Pkg`, and `notarizeUniversalDmg/Pkg` tasks
- Installer filenames get a `_universal` suffix instead of `_arm64`/`_x64`
- Compatible with AOT cache generation (both architectures get their own cache)

## Files changed

| File | Change |
|------|--------|
| `UniversalBinarySettings.kt` | **New** — DSL class (`enabled`, `x64JdkPath`) |
| `AbstractCreateDistributableX64Task.kt` | **New** — Subprocess task launching x64 Gradle build |
| `AbstractMergeUniversalBinaryTask.kt` | **New** — Merges arm64+x64 with lipo, clears xattr, re-signs |
| `PlatformSettings.kt` | Add `universalBinary` block to `JvmMacOSPlatformSettings` |
| `JvmApplicationContext.kt` | Suffix `appTmpDir` with `-x64` in subprocess to isolate intermediates |
| `configureJvmApplication.kt` | Output dir redirect (`-x64`), `configureUniversalBinaryTasks()` |
| `AbstractJPackageTask.kt` | Add `universalBinaryFlag` for `_universal` filename suffix |
| `README.md` | Full documentation of the feature |

## Test plan

- [x] Plugin compiles with JDK 21
- [x] `mergeUniversalBinary` produces a universal `.app` (verified with `lipo -info`)
- [x] `packageUniversalDmg` produces `_universal.dmg` with correct content
- [x] `codesign --verify --deep --strict` passes on the universal app
- [ ] Test on an Intel Mac to confirm the app launches natively


asnaeb:
May I ask why this feature was discarded and if there are any plans to implement it in the future?

kdroidFilter:
I thought it would be simpler to do it on the CI—there’s already a CI setup ready for it—but is it really worth it, considering Apple has just dropped support for Intel Macs?

asnaeb:
> I thought it would be simpler to do it on the CI—there’s already a CI setup ready for it—but is it really worth it, considering Apple has just dropped support for Intel Macs?

In my opinion, it is. 

Not everyone run their CI on github and there are still many users with Intel macs. 

I found myself in this situation: we develop on gitlab and we have users that require Intel mac support. My option would be to tranlsate the github action into gitlab ci's, which are not trivial and need to be mantained and kept in sync with future nucleus updates, or setup my own workflow to provide x64 builds and merge them or deploy them separately. 
Not to mention the need for local multi-arch builds.

I think that - since the work was almost done in this pr - it is worth supporting this. As more users will start to adopt this framework (which is awesome btw) this requirement will rise.

PR #214, opened 2026-04-23
feat(linux-hidpi): detect GNOME fractional scale via Mutter DBus

23rd:
## 🚀 Description

Add a Mutter DBus probe to `linux-hidpi` so `getLinuxNativeScaleFactor()`
returns the correct value on GNOME/Wayland sessions that use fractional
scaling. The probe sits between the existing GSettings step and the
`GDK_SCALE` / `GDK_DPI_SCALE` / `Xft.dpi` fallbacks, uses libgio via
`dlopen` (no new build-time dep), and returns the logical-monitor scale
of the primary display.

## 📄 Motivation and Context

On a 2018 MacBook Pro 15" with Retina panel (2880×1800) running Fedora 43,
GNOME 49 Wayland, GNOME picks **166.67% (5/3)** as the default scale —
the logical desktop is 1728×1080 and every GTK/Qt/Firefox app renders at
that scale. Any Compose Desktop / Swing / AWT app, though, opens its
window at ~1/3 the screen width: AWT renders at 1× because
`sun.java2d.uiScale` is never set.

Tracing [wgtunnel/desktop](https://github.com/wgtunnel/desktop) into
nucleus, I found that `readGnomeScaleFactor()` reads the GSettings
**integer** key `org.gnome.desktop.interface/scaling-factor`, which
stays at `0` on any session using `scale-monitor-framebuffer` —
fractional scaling replaces it, it doesn't populate it. The remaining
fallbacks (`GDK_SCALE`, `GDK_DPI_SCALE`, `Xft.dpi`) are also unset on a
pure Wayland session launched from a `.desktop` entry, so the detector
ultimately returns `0.0`.

The only process that knows the current fractional scale is Mutter; it
exposes it via `org.gnome.Mutter.DisplayConfig.GetCurrentState`. This PR
consults that interface before giving up.

Tested on the setup above (Fedora 43, GNOME 49 Wayland, Mesa 26.0.4,
`scale-monitor-framebuffer` enabled).

**Direct JNI probe of the patched .so:**

| Env                        | Returned |
|----------------------------|---------:|
| default (Mutter reachable) | `1.6666666269302368` |
| `J2D_UISCALE=3`            | `3.0` (explicit override still wins) |
| `GDK_SCALE=2`              | `1.6666666269302368` (Mutter beats GDK, as intended) |

Value matches `gdbus call --session --dest org.gnome.Mutter.DisplayConfig
--object-path /org/gnome/Mutter/DisplayConfig --method
org.gnome.Mutter.DisplayConfig.GetCurrentState` exactly.

**End-to-end via a downstream Compose app:**

Published this branch to `mavenLocal` and pointed an unmodified
`wgtunnel/desktop` checkout at it:

```
Info: Applied Linux HiDPI scale: 1.6666666269302368
sun.java2d.uiScale = 1.6666666269302368
```

Window now opens at the correct physical size — same visual footprint
as Firefox / Nautilus / Settings. Before the patch the same setup
rendered at 1× and looked ~1/3 the expected width.

**Multi-monitor:**

Also verified side-by-side with a 1920×1080 external monitor
(logical-monitors: `(0,0, 1.6666…, primary, eDP-1)` +
`(1728,0, 1.0, secondary, DP-4)`). The probe picks the primary entry,
so `sun.java2d.uiScale` is set to `1.6666…` for the whole process.
AWT has no per-monitor DPI in JDK 21, so dragging between monitors
keeps the same scale — usable and readable on both panels, no
crash / Y-flip / input misalignment.

**Project checks:**

- `./gradlew :linux-hidpi:build` — compile + ktlint clean.

## 📦 Types of changes

## ✅ Checklist

   Mirrors the existing `readGnomeScaleFactor()` pattern in the same
   file: `dlopen("libgio-2.0.so.0", RTLD_LAZY | RTLD_LOCAL)`,
   `dlsym`-ed function pointers, GError / GVariant / GObject refcount
   balance, any failure returns `0.0` so the remaining fallback chain
   still runs. No new hard link-time dependency.
   `LinuxHiDpi.kt` KDoc now lists Mutter in the priority order.

### DIFF
--- a/linux-hidpi/src/main/kotlin/io/github/kdroidfilter/nucleus/hidpi/LinuxHiDpi.kt
+++ b/linux-hidpi/src/main/kotlin/io/github/kdroidfilter/nucleus/hidpi/LinuxHiDpi.kt
@@ -7,9 +7,10 @@ package io.github.kdroidfilter.nucleus.hidpi
  * Sources consulted in priority order:
  *   1. `J2D_UISCALE`   — explicit JVM override (env var)
  *   2. GSettings       — GNOME `org.gnome.desktop.interface` → `scaling-factor`
- *   3. `GDK_SCALE`     — GTK / GNOME session variable
- *   4. `GDK_DPI_SCALE` — GTK fractional DPI multiplier
- *   5. `Xft.dpi`       — X Resource Manager (KDE, legacy GNOME, …)
+ *   3. Mutter DBus     — GNOME fractional scale via `org.gnome.Mutter.DisplayConfig`
+ *   4. `GDK_SCALE`     — GTK / GNOME session variable
+ *   5. `GDK_DPI_SCALE` — GTK fractional DPI multiplier
+ *   6. `Xft.dpi`       — X Resource Manager (KDE, legacy GNOME, …)
  *
  * @return A positive scale factor (e.g. `2.0` for a 200 % HiDPI display),
  *         or `0.0` when the scale cannot be determined (let the JVM decide).
--- a/linux-hidpi/src/main/native/linux/nucleus_hidpi_linux.c
+++ b/linux-hidpi/src/main/native/linux/nucleus_hidpi_linux.c
@@ -9,9 +9,10 @@
  * Detection order (same priority as JBR):
  *   1. J2D_UISCALE   — explicit JVM override (env var)
  *   2. GSettings     — GNOME integer scaling via libgio (dlopen, no hard dep)
- *   3. GDK_SCALE     — GTK environment variable
- *   4. GDK_DPI_SCALE — GTK fractional DPI multiplier
- *   5. Xft.dpi       — X Resource Manager via libX11 (dlopen, no hard dep)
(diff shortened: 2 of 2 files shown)


kdroidFilter:
Ah yes, I copied the JBR implementation, but it's much cleaner to use D-Bus! Thank you very much, I'll take a look at that.

kdroidFilter:
The 2-second DBus timeout bothers me; it's really necessary ?

23rd:
I'll check tomorrow.

23rd:
In the happy path `GetCurrentState` returns well under 100 ms, so 2 s vs the GDBus 25 s default is invisible. The bound only matters in the pathological case (mutter mid-restart, wedged session bus), because the probe runs synchronously before AWT init while the main UI is being built.

If a tighter bound is preferred, 500 ms works; or we can pass `-1` to inherit the GDBus default:

```c
-1   /* GDBus default timeout */,
```

Issue #129, opened 2026-03-20
Support full-window content layouts with decorated title bar


HansHolz09:
## ⚠️ Is your feature request related to a problem? Please describe
At the moment, `DecoratedWindow` always lays out the main Compose content below the title bar area.

This makes it impossible to build layouts where the window content continues behind or underneath the title bar, while still keeping the decorated window behavior and native window controls where applicable.

## 💡 Describe the solution you'd like
I would like Nucleus Decorated Window to support a mode where the Compose content can continue drawing underneath the title bar area.

The exact implementation is not important as long as the following becomes possible:
- the main window content can extend through the full window height, including the area behind or under the title bar
- native window controls and platform-specific behavior continue to work correctly
- dragging, hit testing, control button placement, and content insets remain handled correctly

In other words, this request is about enabling full-window content layouts without losing the existing decorated-window behavior.

## 🤚 Do you want to develop this feature yourself?

- [x] Yes
- [ ] No

PR #245, opened 2026-05-15
feat(window): add fill-center title bar policy and BasicTitleBar APIs

RavenLiao:
## 🚀 Description
This PR introduces a new title bar layout customization layer with a built-in fill-center policy and adds dedicated basic title bar APIs for advanced layout scenarios.

Key updates include:
- Added `TitleBarLayoutPolicy` with `Default` (legacy-compatible) and `FillCenter` behavior.
- Added `BasicTitleBar(...)` and `BasicDialogTitleBar(...)` as low-level entry points.
- Kept existing `TitleBar(...)` / `DialogTitleBar(...)` APIs compatible by delegating to `Basic*` + `Default`.
- Wired `layoutPolicy` through core/JBR/JNI title bar paths.
- Added Fill Title demo window in `example` for behavior validation.
- Updated demo tab interactions in title bar to use `titleBarClickable`, improving macOS fullscreen interaction stability.

## 📄 Motivation and Context
Current title bar center content behavior is primarily wrap-content oriented, which makes "center fills remaining space" scenarios (e.g. tab strips) hard to express consistently.

This change provides a framework-native fill-center layout model while preserving existing behavior and API compatibility:
- No breaking changes for existing title bar usage.
- Advanced users can opt into `FillCenter` via `Basic*` APIs.
- Platform integration details remain internal; only content layout policy is exposed.

Build/test checks:
- `./gradlew.bat :decorated-window-core:compileKotlin :decorated-window-jbr:compileKotlin :decorated-window-jni:compileKotlin :example:compileKotlin`  
  Result: **BUILD SUCCESSFUL**

Manual validation:
- Verified Fill Title demo behavior in windowed mode and fullscreen mode.
- Confirmed title bar controls and content still render/behave correctly with new policy plumbing.
- Verified macOS fullscreen demo tab switching no longer relies on plain `clickable` in title bar hit-test-sensitive area.

## 📦 Types of changes

## ✅ Checklist

### DIFF
--- a/decorated-window-core/src/main/kotlin/io/github/kdroidfilter/nucleus/window/DialogTitleBarImpl.kt
+++ b/decorated-window-core/src/main/kotlin/io/github/kdroidfilter/nucleus/window/DialogTitleBarImpl.kt
@@ -17,6 +17,7 @@ fun DecoratedDialogScope.DialogTitleBarImpl(
     gradientStartColor: Color = Color.Unspecified,
     style: TitleBarStyle = LocalTitleBarStyle.current,
     controlButtonsDirection: LayoutDirection = LocalLayoutDirection.current,
+    layoutPolicy: TitleBarLayoutPolicy = TitleBarLayoutPolicy.Default,
     applyTitleBar: (Dp, DecoratedWindowState) -> PaddingValues,
     onPlace: (() -> Unit)? = null,
     backgroundContent: @Composable () -> Unit = {},
@@ -30,10 +31,36 @@ fun DecoratedDialogScope.DialogTitleBarImpl(
         gradientStartColor = gradientStartColor,
         style = style,
         controlButtonsDirection = controlButtonsDirection,
+        layoutPolicy = layoutPolicy,
         applyTitleBar = applyTitleBar,
         onPlace = onPlace,
         backgroundContent = backgroundContent,
     ) { _ ->
         content(dialogState)
     }
+}
+
+@Suppress("FunctionNaming", "LongParameterList")
+@Composable
+fun DecoratedDialogScope.DialogTitleBarImpl(
+    modifier: Modifier = Modifier,
+    gradientStartColor: Color = Color.Unspecified,
+    style: TitleBarStyle = LocalTitleBarStyle.current,
+    controlButtonsDirection: LayoutDirection = LocalLayoutDirection.current,
+    applyTitleBar: (Dp, DecoratedWindowState) -> PaddingValues,
+    onPlace: (() -> Unit)? = null,
+    backgroundContent: @Composable () -> Unit = {},
+    content: @Composable TitleBarScope.(DecoratedDialogState) -> Unit,
+) {
+    DialogTitleBarImpl(
+        modifier = modifier,
+        gradientStartColor = gradientStartColor,
+        style = style,
+        controlButtonsDirection = controlButtonsDirection,
+        layoutPolicy = TitleBarLayoutPolicy.Default,
+        applyTitleBar = applyTitleBar,
+        onPlace = onPlace,
+        backgroundContent = backgroundContent,
+        content = content,
+    )
 }
--- a/decorated-window-core/src/main/kotlin/io/github/kdroidfilter/nucleus/window/TitleBarCore.kt
+++ b/decorated-window-core/src/main/kotlin/io/github/kdroidfilter/nucleus/window/TitleBarCore.kt
@@ -24,30 +24,23 @@ import androidx.compose.ui.input.pointer.PointerEventPass
 import androidx.compose.ui.input.pointer.PointerEventType
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.layout.Layout
-import androidx.compose.ui.layout.Measurable
-import androidx.compose.ui.layout.MeasurePolicy
-import androidx.compose.ui.layout.MeasureResult
-import androidx.compose.ui.layout.MeasureScope
-import androidx.compose.ui.layout.Placeable
 import androidx.compose.ui.layout.layoutId
+import androidx.compose.ui.layout.onPlaced
 import androidx.compose.ui.layout.onSizeChanged
 import androidx.compose.ui.node.ModifierNodeElement
 import androidx.compose.ui.node.ParentDataModifierNode
 import androidx.compose.ui.platform.InspectorInfo
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalLayoutDirection
-import androidx.compose.ui.unit.Constraints
 import androidx.compose.ui.unit.Density
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.LayoutDirection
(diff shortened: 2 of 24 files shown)


kdroidFilter:
Thank you very much for your work, I'm thinking of accepting the PR, although long-term I wonder if using a title bar is really a good pattern. It's originally a fork of Jewel, but honestly I think we'd just need a modifier that indicates a given composable should allow the window to be dragged, and simply overlay the native buttons in the right position — that would allow for unified UIs with a sidebar and not necessarily a title bar. This would also resolve issue https://github.com/NucleusFramework/Nucleus/issues/129 and would perhaps make things much simpler. I don't think I'd deprecate the title bar, but in my opinion it's just an old inherited pattern

RavenLiao:
> Thank you very much for your work, I'm thinking of accepting the PR, although long-term I wonder if using a title bar is really a good pattern. It's originally a fork of Jewel, but honestly I think we'd just need a modifier that indicates a given composable should allow the window to be dragged, and simply overlay the native buttons in the right position — that would allow for unified UIs with a sidebar and not necessarily a title bar. This would also resolve issue #129 and would perhaps make things much simpler. I don't think I'd deprecate the title bar, but in my opinion it's just an old inherited pattern


Yes, something like this?

```kotlin
data class WindowControlsArea(
    val side: WindowControlsSide, // Start (macOS) or End (Windows)
    val width: Dp,                // horizontal space the buttons occupy
    val height: Dp,               // how far down they extend
    val offsetY: Dp = 0.dp,       // shifts during fullscreen menu-bar animation
)

val LocalWindowControlsArea = staticCompositionLocalOf { WindowControlsArea(...) }

// ---

DecoratedWindow(...) {
    val controls = LocalWindowControlsArea.current

    Box(Modifier.fillMaxSize()) {
        HeroImage(Modifier.fillMaxSize())

        Sidebar(
            Modifier
                .windowDraggableArea()
                .windowControlsPadding() // reads LocalWindowControlsArea internally
        )
    }
}
```

`windowControlsPadding()` is just a convenience on top of `LocalWindowControlsArea`; code that needs precise avoidance (e.g. clipping behind the traffic lights) can read the area directly. 

This also feels like the natural resolution to #129 — content extends freely, each element opts in to avoidance rather than being forced into a fixed TitleBar layout.

kdroidFilter:
Yes, the direction looks right. One thing I'd want baked in from day one: the ability to force a specific OS look regardless of the host. A developer on Linux should be able to display the macOS traffic lights for previews/screenshots, with `DecoratedWindow` drawing the buttons in Compose.

Something like:

```kotlin
DecoratedWindow(
    controlsStyle = WindowControlsStyle.Native, // default: real OS chrome
    // WindowControlsStyle.MacOS / Windows / Linux forces Compose-drawn controls
)
```

RavenLiao:
> Yes, the direction looks right. One thing I'd want baked in from day one: the ability to force a specific OS look regardless of the host. A developer on Linux should be able to display the macOS traffic lights for previews/screenshots, with `DecoratedWindow` drawing the buttons in Compose.
> 
> Something like:
> 
> ```kotlin
> DecoratedWindow(
>     controlsStyle = WindowControlsStyle.Native, // default: real OS chrome
>     // WindowControlsStyle.MacOS / Windows / Linux forces Compose-drawn controls
> )
> ```

I explored this direction a bit further, and at this point I’m leaning toward the following:

Instead of continuing to extend the existing `DecoratedWindow` / `TitleBar` model, I think it would make more sense to introduce a separate API, tentatively called `CustomChromeWindow`.

What I want from it is fairly straightforward:

- a full-content window
- explicit window controls
- the ability to force a specific OS controls appearance for previews / screenshots
- no compatibility impact on the existing `DecoratedWindow`

The API shape I currently have in mind is roughly:

```kotlin
@Stable
interface CustomChromeWindowScope : FrameWindowScope {
    val state: DecoratedWindowState
    val controlsStyle: WindowControlsStyle
}

@Composable
fun CustomChromeWindow(
    onCloseRequest: () -> Unit,
    state: WindowState = rememberWindowState(),
    visible: Boolean = true,
    title: String = "",
    icon: Painter? = null,
    resizable: Boolean = true,
    enabled: Boolean = true,
    focusable: Boolean = true,
    alwaysOnTop: Boolean = false,
    minimumSize: DpSize? = null,
    controlsStyle: WindowControlsStyle = WindowControlsStyle.Native,
    content: @Composable CustomChromeWindowScope.() -> Unit,
)
```

```kotlin
sealed interface WindowControlsStyle {
    data object Native : WindowControlsStyle

    data class Compose(
        val appearance: WindowControlsAppearance = WindowControlsAppearance.Host,
    ) : WindowControlsStyle
}

enum class WindowControlsAppearance {
    Host,
    MacOS,
    Windows,
    Linux,
}
```

```kotlin
@Composable
fun CustomChromeWindowScope.WindowControls(
    modifier: Modifier = Modifier,
)
```

```kotlin
@Immutable
data class WindowControlsArea(
    val side: WindowControlsSide,
    val width: Dp,
    val height: Dp,
    val offsetY: Dp = 0.dp,
    val safePadding: PaddingValues = PaddingValues(0.dp),
)

val LocalWindowControlsArea =
    staticCompositionLocalOf<WindowControlsArea?> { null }

fun Modifier.windowControlsPadding(): Modifier
```

Intended usage would look something like:

```kotlin
CustomChromeWindow(
    controlsStyle = WindowControlsStyle.Native,
) {
    WindowControls()
    // custom full-content layout
}
```

And if you want to force a specific platform style for preview purposes:

```kotlin
CustomChromeWindow(
    controlsStyle = WindowControlsStyle.Compose(WindowControlsAppearance.MacOS),
) {
    WindowControls()
}
```

My current view is:

- this should be a new window-level API, rather than adding more semantics onto `TitleBar`
- drag regions are probably not the main problem here, since Compose Desktop already has `WindowDraggableArea()`, which seems sufficient for most desktop use cases

So this proposal feels more about adding these capabilities:

- an explicit controls primitive like `WindowControls()`
- controls occupancy / safe-area information like `WindowControlsArea`
- Compose-rendered macOS controls
- a more general chrome-region / hit-test registry
- a more general top-chrome metrics / native integration abstraction

What I’m not sure about is which of these should stay purely at the frontend API / composition layer, and which are worth expanding backend integration for.

My impression is that these items:

- `WindowControls()`
- `WindowControlsArea`
- Compose-rendered macOS controls

look like things that could be pushed forward at the frontend layer first.

The two parts that seem more likely to require backend work are:

- a more general chrome-region / hit-test registry
- a more general top-chrome metrics / native integration abstraction

Mainly because if we go in that direction, the current backend integration still looks fairly `title-bar-first`:

- JBR-side hit testing / client regions are still centered around `TitleBarInfo`
- Windows JNI is still centered around a top height band
- macOS JNI is still centered around title bar height synchronization

So the main thing I’d like to get feedback on is:

- whether it makes sense to expand backend integration for those two more general capabilities
- or whether it would be better to keep the new API scoped to the frontend layer first and leave the backend as-is

If this direction makes sense, I can continue refining it into a more concrete API sketch and implementation plan.

PR #254, opened 2026-05-29
fix(plugin): publish one merged S3 auto-update manifest per OS

### LINKED ISSUE #253: Multiple target formats for the same OS fails on S3
## 🐛 Describe the bug

After building, the beta-linux.yml only contains info for the .deb when deb and appimage formats are specified.

## ⚠️ Current behavior

I believe the builder is writing 2 files that race to be written and the last one wins.

sproctor:
## 🚀 Description

Claude generated description of changes:

Nucleus runs a separate electron-builder invocation per TargetFormat, and each updatable run publishes its own single-artifact `<channel><osSuffix>.yml`. Because the manifest name is keyed only on the OS, every format of the same OS publishes to the same S3 key, so the last writer clobbers the others — e.g. the deb run overwrites the AppImage entry, and an AppImage client then hits NoMatchingFileException (on macOS a dmg-only `latest-mac.yml` breaks Squirrel.Mac updates entirely).

Fix: always set `publishAutoUpdate: false` for S3 so electron-builder uploads the artifacts but not the manifests, then a new merge task discovers the per-format manifests for the current OS, unions their `files:` entries (UpdateYmlMerger), and uploads a single manifest via `aws s3 cp`. A lone manifest passes through verbatim to preserve electron-builder's exact metadata.

- TargetFormat.producesUpdateManifest gates registering the merge task.
- Upload mirrors electron-builder's publish decision (publish mode + CI tag), fails loudly if a manifest is expected but missing or the aws CLI is absent.
- Tested: merge/discovery, publish-flag/tag logic, s3 key, and the publishAutoUpdate:false config emission.

Scope: S3 only. GitHub releases are already consolidated in CI (generate-update-yml).

## 📄 Motivation and Context

Fixes #253

Added tests.

## 📦 Types of changes

## ✅ Checklist

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,7 @@
 [versions]
 agp = "8.13.2"
 asm = "9.9.1"
+awsSdk = "2.45.1"
 batik = "1.19"
 coilVersion = "3.4.0"
 compose = "1.10.3"
@@ -80,4 +81,6 @@ compose-material-icons-extended = { module = "org.jetbrains.compose.material:mat
 reorderable = { module = "sh.calvin.reorderable:reorderable", version.ref = "reorderable" }
 thumbnailator = { module = "net.coobird:thumbnailator", version.ref = "thumbnailator" }
 asm = { module = "org.ow2.asm:asm", version.ref = "asm" }
+aws-s3 = { module = "software.amazon.awssdk:s3", version.ref = "awsSdk" }
+aws-url-connection-client = { module = "software.amazon.awssdk:url-connection-client", version.ref = "awsSdk" }
 kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "kotlinxSerialization" }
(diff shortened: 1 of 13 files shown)


kdroidFilter:
Thank you very much for your contribution, I think it would be necessary to document this, could you add it to the documentation?

sproctor:
I'm not sure what there is to document here. This is just a straight bug fix. Before electron-builder was uploading the yml files, one per platform that would overwrite other target packages. Now the yml files are merged for each platform. While thinking about it a bit further, I think it would be simpler to just have one yml file per package type. latest-msi.yml, beta-msi.yml, latest-exe.yml, etc. It could go back to being electron-builder's responsibility to upload the files. It would require some changes to the build system and updater though.

Edit: You're right, I forgot about the aws client requirement. I'll see if I can remove that instead.

sproctor:
Okay, should no longer need any documentation update. Switched to using AWS library, so it doesn't need the aws client to be pre-installed.

PR #285, opened 2026-07-05
refactor(tao): extract ModalInputBlocker to fix detekt complexity

kdroidFilter:
## Summary
The Maven publish workflow's publish step is gated behind pre-merge checks; `openDecoratedWindowLinux` trips detekt `CyclomaticComplexMethod` (21 > 20), so pre-merge fails and the publish is skipped.

Extract the modal-dialog input blocker / scrim (duplicated in the Linux and Windows scene setups) into a shared `ModalInputBlocker` composable, moving its `if`/`||`/`&&`/`while` out of both host functions and dropping their cyclomatic complexity below the limit. No behavior change.

## Test plan
- [x] `:decorated-window-tao:compileKotlin`, `ktlintCheck`, `detekt` all clean locally
- [ ] CI Pre Merge Checks green (gating the publish)


kdroidFilter:
Not the publish blocker — decorated-window-tao is not part of the preMerge gate, so its detekt isn't run in CI. The real cause was updater-runtime SwallowedException (#286). Closing; the tao complexity can be a separate cleanup if we ever add tao to preMerge.

PR #296, opened 2026-07-10
ci(release-desktop): download native artifacts before packaging sample

kdroidFilter:
## Summary\n\n- Add a dependency on the reusable  workflow so all platform native libraries are built before packaging.\n- Download every native artifact into the correct  directory before running .\n- Aligns  with , , and .\n- Fixes packaged example app crashes on Windows caused by missing ANGLE DLLs (, ) from .\n\n## Test plan\n\n- [ ] Run  on Windows.\n- [ ] Verify the installed  contains  and .\n- [ ] Launch the packaged app and confirm it reaches the main window without .\n


kdroidFilter:
Recreating with the correct base branch (nucleus-2.0) from a dedicated fix branch.

PR #301, opened 2026-07-11
feat(tao): standalone transparent popup panel + single-instance restore hook

kdroidFilter:
## What

Infrastructure for tray-style applications on the Tao backend: a popup that renders in a **top-level, ownerless, per-pixel transparent native panel** — no backing window (nothing in taskbar / Alt-Tab / task view), no host render loop. Built for and consumed by ComposeNativeTray's `TrayApp`.

## New public API

- **`TaoStandalonePopup`** (Windows, no-op elsewhere): visibility, absolute dp positioning, focusability, outside-click + key handlers; bridges the caller's composition locals into the panel's scene (reactively — dark-mode switches propagate).
- **`TaoScreenGeometry`**: primary-monitor work area / scale factor without needing a `TaoWindow`.
- **`DecoratedWindow(nativePopupLayers = true)`**: opt-in `PlatformLayersComposeScene` so regular Compose `Popup`s materialize as native transparent panels.
- **`SingleInstanceRestoreEffect`** (nucleus-application): fires when a second instance attempts to launch — for apps whose main surface is not a decorated window (tray apps re-opening their popup).

## Implementation notes

- **Rendering**: ownerless `WS_POPUP` + `NOACTIVATE|TOOLWINDOW|NOREDIRECTIONBITMAP|TOPMOST`, DComp premultiplied alpha, own `CanvasLayersComposeScene`, on-demand rendering on the Tao main thread. Pacing runs on an absolute 60 fps grid with evenly spaced frame-clock timestamps; `timeBeginPeriod(1)` (refcounted) only while a panel is visible — without it the JVM scheduler's ~15.6 ms quantum halves the frame rate.
- **Headless EGL bootstrap** (`nativeEnsureHeadlessContext`): creates the shared ANGLE context before any window host exists, so tray-only apps work.
- **Outside-click**: global `WH_MOUSE_LL` for ownerless panels (a thread-local `WH_MOUSE` never sees clicks on other apps). The LL hook never calls into Java — it `PostMessage`s to the panel and the WndProc does the JNI upcall (LowLevelHooksTimeout safety). Hook refcounts count requests and retry on failed installs.
- **Keyboard**: focus acquisition on click and on show (`GetForegroundWindow` comparison + `AttachThreadInput` to bypass the foreground lock), AltGr filtered as text input (mirrors tao's `filter_out_altgr`), surrogate-pair code points.
- **Key dispatch factored** into `ComposeScene.dispatchNativeKeyEvent` (used by both popup layers, the macOS overlay and the standalone panel). Fixes `keyLocation = UNKNOWN` making `Key.*` constants (Backspace/Enter/arrows) unmatchable, and adds the missing synthetic KEY_TYPED to the Windows popup-layer path.
- **GraalVM**: reachability metadata for the new JNI callbacks (named inner classes).

## Testing

- `StandalonePanelNativeSmokeTest`: headless EGL bootstrap → ownerless panel → EGL make-current → Skia DirectContext, without a Tao event loop.
- Validated end-to-end by ComposeNativeTray's TrayApp demo on Windows 11 (JVM + GraalVM native image): display, animations, outside-click on other apps/desktop, text input incl. AZERTY AltGr, Backspace/arrows, I-beam cursor, topmost.

### DIFF
--- a/decorated-window-material2/src/main/kotlin/dev/nucleusframework/window/material2/MaterialDecoratedWindow.kt
+++ b/decorated-window-material2/src/main/kotlin/dev/nucleusframework/window/material2/MaterialDecoratedWindow.kt
@@ -77,6 +77,10 @@ fun NucleusApplicationScope.MaterialDecoratedWindow(
     enabled: Boolean = true,
     focusable: Boolean = true,
     alwaysOnTop: Boolean = false,
+    // Materialise Compose Popup layers as native transparent windows
+    // (NSPanel / WS_POPUP HWND) so menus can extend past the window bounds.
+    // Honoured by the Tao backend on Windows/macOS; ignored elsewhere.
+    nativePopupLayers: Boolean = false,
     minimumSize: DpSize? = null,
     onPreviewKeyEvent: (KeyEvent) -> Boolean = { false },
     onKeyEvent: (KeyEvent) -> Boolean = { false },
@@ -105,6 +109,7 @@ fun NucleusApplicationScope.MaterialDecoratedWindow(
             enabled = enabled,
             focusable = focusable,
             alwaysOnTop = alwaysOnTop,
+            nativePopupLayers = nativePopupLayers,
             minimumSize = minimumSize,
             onPreviewKeyEvent = onPreviewKeyEvent,
             onKeyEvent = onKeyEvent,
--- a/decorated-window-material3/src/main/kotlin/dev/nucleusframework/window/material/MaterialDecoratedWindow.kt
+++ b/decorated-window-material3/src/main/kotlin/dev/nucleusframework/window/material/MaterialDecoratedWindow.kt
@@ -90,6 +90,10 @@ fun NucleusApplicationScope.MaterialDecoratedWindow(
     enabled: Boolean = true,
     focusable: Boolean = true,
     alwaysOnTop: Boolean = false,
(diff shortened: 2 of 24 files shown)


kdroidFilter:
**Pre-merge visual checks** (the only two code paths whose observable behavior changes for existing users):

1. **Overlay popup clamping** — `TaoPopupSceneLayerWindows` now sizes its scene against the monitor *work area* instead of the parent window. A `DropdownMenu` inside a `NativeView` overlay near a window edge used to clamp/flip within the window bounds; it can now legitimately extend past them (macOS contract alignment — intended). Verify one existing overlay popup near a window edge on Windows.

2. **Esc-to-dismiss on popup layers** — the `keyLocation` fix makes non-character keys (Esc, Enter, arrows, Backspace) matchable by popup content for the first time. The dismiss-on-Esc handler is the layer's *fallback* `onKeyEvent` (standard `ComposeSceneLayer.setKeyEventListener` contract, same as the AWT backend): if popup content consumes Esc, dismiss won't fire — previously it always fired only because content could never match Esc. Verify Esc still closes a plain `DropdownMenu` on the Windows and macOS layer paths.

**Known accepted cost**: `attachedHostCount` now includes standalone panels, so while a tray panel exists (even hidden) window hosts run in `resetGLAll`-per-frame mode. Required by the shared EGL context; window-only apps are unaffected.

PR #345, opened 2026-07-14
feat(graalvm): reduce native image binary size

kdroidFilter:
## Summary

GraalVM native images produced by Nucleus were large (~135 MB) because no size-oriented settings were applied: `buildArgs` defaulted to empty (so native-image built with the speed-oriented `-O2`) and only companion `.so`/`.dylib` files were stripped — never the main executable.

- **New DSL flag `graalvm { optimizeForSize = true }`** — injects `-Os` (optimize for binary size). Default `false` (opt-in, since `-Os` slightly favors size over runtime speed). Placed before user `buildArgs` so an explicit `-O*` still takes precedence. Typically trims ~20–30% off Compose images.
- **Strip the main native-image executable** (always on, consistent with the existing companion-library stripping):
  - **Linux**: `strip` the executable after `patchelf` — removes the `.symtab` (native-image emits no separate debug info without `-g`), reclaiming tens of MB.
  - **macOS**: `strip -x` after the vtool/rpath edits and before the ad-hoc `codesign` (which re-signs the bundle). `-x` keeps external/global symbols required by the flat-namespace AWT dylibs and the cursor stub.
  - **Windows**: unchanged — PE/COFF carries no `.symtab` and native-image emits no `.pdb` without `-g`, so there is nothing to strip.

## Test plan

- [x] `:plugin:compileKotlin` succeeds
- [ ] Linux native build: verify executable shrinks and still launches
- [ ] macOS native build: verify `strip -x` + `codesign` succeed and the `.app` launches
- [ ] Verify `optimizeForSize = true` produces a smaller binary than the default

### DIFF
--- a/plugin-build/plugin/src/main/kotlin/dev/nucleusframework/desktop/application/dsl/GraalvmSettings.kt
+++ b/plugin-build/plugin/src/main/kotlin/dev/nucleusframework/desktop/application/dsl/GraalvmSettings.kt
@@ -29,6 +29,14 @@ abstract class GraalvmSettings
         // must run on any x86-64 CPU. "native" optimizes for the build machine only and crashes on
         // older CPUs ("does not support all of the following CPU features"). Override for local perf.
         val march: Property<String> = objects.notNullProperty("compatibility")
+
+        // Optimize the native image for binary size (`-Os`) instead of the default speed-oriented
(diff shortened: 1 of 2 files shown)


kdroidFilter:
Note: this branch was split. The macOS DMG LZMA recompression that was previously bundled here has moved to #346. This PR is now scoped to the native-image binary-size reduction only (`-Os` + strip).

PR #373, opened 2026-07-19
test(decorated-window-tao): tiered test infra — offscreen scene harness + headful window suite

kdroidFilter:
## Summary

Full test infrastructure for the Tao backend (stacked on #371) — every layer verified e2e in this PR's CI:

**Stage 1 — offscreen scene harness** (`TaoSceneTestHarness`, 60 cases): same `CanvasLayersComposeScene` config as the host, production `TaoWindowInfo`, thread-safe manually-pumped dispatcher, owned frame clock, frames recorded via the production `recordSceneToPicture` CPU path with virtual time. Input through the production translation layer, host input-guards mirrored. Pixel + semantics assertions.

**Stage 2 — headful window suite** (`taoHeadfulTest`, 6 cases): real Tao windows, one event loop per process (`runApplicationTest` model), drivers at application level. **Found and fixed a real native bug**: macOS `setMaximized(false)` issued during the zoom animation was a silent no-op (vendored-tao PATCH: dedup on tracked state, standard_frame saved only when maximizing); the case now regression-tests it by unmaximizing mid-animation. Linux runs all 6 cases under Xvfb+openbox.

**GraalVM native pyramid** (`examples/tao-native-test` + `tao-native-tests` job in test-graalvm.yaml): the whole pyramid compiled INTO a native image and executed per OS (battery 60/60 + headful 6/6, `native=true`) — the first place a produced native binary is actually run in CI. Reflection-free registry; the binary must be the PACKAGED one (skiko/jawt/icudtl extracted next to it).

**Native accessibility protocol** (`scripts/ci/verify-{uia.ps1,atspi.py,ax.swift}` + `tao-a11y` job): external OS-level probes against tao-demo's A11y tab (`NUCLEUS_DEMO_TAB`) — the same client APIs assistive tech uses. **macOS: hosted runners are TCC-trusted → 4/4 AX assertions PASS in CI.** Linux: AT-SPI exposure + doAction round-trip PASS. Windows: full protocol (UIA raw-view walk, Invoke/Toggle/RangeValue round-trips + advanced coverage) — **hard-passes 10/10 on hosted runners** since the idle-activation fix (the parked semantics walk re-arms when an AT connects; the empty tree was never a GL/display limitation — the Windows host renders via ANGLE/D3D11).

**CI fixes along the way**: `pull_request.branches: '*'` silently skipped every workflow for PRs based on `refactor/*` (fixed to `'**'` in pre-merge, test-graalvm, test-packaging); `decorated-window-tao:check` joins `preMerge`; harness dispatcher made thread-safe (root cause of all observed flakes).

Windows-agent follow-ups: idle-activation race fixed on all three hosts (AT connecting against a settled window got the 1-node seed tree forever), UIA RangeValue.SetValue now dispatches an absolute set (was a single relative step), harness dispatcher virtualizes delay/withTimeout (root cause of the scroll-residue flake), Mesa CI step removed as irrelevant.

## CI (this PR, all green)
- gradle (preMerge) · tao-tests ×3 OS · tao-headful ×3 OS · tao-a11y ×3 OS · tao-native-tests ×3 OS


kdroidFilter:
## Windows UIA protocol validated on real hardware — and the hosted-runner mystery solved

Ran the full stack on a physical Windows 11 machine with a real GPU:

| Suite | Result |
|---|---|
| `:decorated-window-tao:test` (offscreen battery) | 60/60 |
| `:decorated-window-tao:taoHeadfulTest` | 6/6 |
| GraalVM native runner (`tao-native-test.exe battery` / `headful`, packaged binary) | 60/60 + 6/6 |
| `scripts/ci/verify-uia.ps1` against tao-demo | **10/10 PASS, 0 SKIPPED** (after the fixes below) |

### The empty-scene hypothesis was wrong

The scene was never empty. On real hardware the window renders perfectly (verified by `PrintWindow` capture — the A11y tab is fully composed), and the Windows host renders through **ANGLE/D3D11** (WARP-capable on VMs), never through WGL/opengl32 — so the Mesa/llvmpipe install on the runners was a no-op by construction. `first pushSnapshot: nodes=1` is normal everywhere: it's the forced seed push that fires before the first composition produces semantics. The debug trace now also logs the first push carrying a real tree (73 nodes on tao-demo).

### The real bug: idle-activation race (reproduced locally)

Probing **right after launch** → all assertions pass. Probing **after ~60 s of idle** → the probe sees only the 9 default non-client elements and everything fails — exactly the CI symptom. Root cause: a UIA client's `WM_GETOBJECT` sets the native active+resync flags, but the debounced semantics walk had already fired and been gated off (`shouldRunSync` false), and nothing re-scheduled it. On a static scene no further semantics change ever comes, so the provider serves the 1-node seed tree forever.

**Fix** (`5cbbf1eb`): the a11y sync scheduler now parks the gated walk and polls the activation gate at 250 ms (one JNI flag read per tick, no render-loop wakeups). The tree is projected within ~250 ms of the first AT query. Applied to all three hosts (the same race exists on macOS/Linux in principle).

Verified: launch → 75 s idle → probe = 10/10 PASS.

### Bonus fix found by the probe

`RangeValue.SetValue(0.7)` from 0.5 landed on 0.6: the C provider translated SetValue into a single ±10% Increment/Decrement step. `a670ffa9` adds a proper absolute set-value channel (C export + Rust trampoline → `dispatchA11ySetValue` → SetProgress with clamping, same path as Linux AT-SPI). Now lands on 0.7 exactly.

### CI changes

`b05152f6` removes the Mesa step from the `tao-a11y` Windows leg and rewrites the job comments. With the activation fix the Windows probe should now run its assertions on hosted runners instead of self-skipping — the self-skip stays as a safety valve.

`check` + `apiCheck` green locally (JBR 21, natives rebuilt via `build.bat`, loader cache purged).

kdroidFilter:
Follow-up: the `tao-tests (windows-latest)` failure on the first CI round was the documented wheel-scroll flake (rare 8px residue), confirmed by a green rerun of the same commit. Root cause: the offscreen harness's `QueueDispatcher` didn't implement `Delay`, so `delay`/`withTimeout` inside scene code (MouseWheelScrollingLogic's coalescing timeout) ran on kotlinx's real-time DefaultExecutor while the harness advanced a virtual clock — a cross-thread race that only bites on loaded runners. `32f229a1` virtualizes those timers onto the harness clock and makes `frameUntilIdle` hold off while a near-horizon timer is pending.

Final state on `32f229a1`: **Pre Merge Checks ✓ (all 9 tao jobs, UIA probe 10/10 on the hosted Windows runner — no Mesa, no self-skip), Test GraalVM Native Image ✓, Test Packaging ✓.**

Issue #394, opened 2026-07-23
MacOS packaging: Signing, bundling and shipping .appex network extension MacOS target together with the main JVM app.


tomasjablonskis:
Lets say we wanted to create a MacOS application that uses Apples Network Extension framework to extend core networking capabilities such as: proxying traffic, managing VPN configurations, managing some Wifi/hotspot settings, filtering content, etc.

<hr/>

In normal MacOS Swift applications Network Extension can be included by configuring a separate Network Extension target with its own signing, entitlements, plist files, etc. When main MacOS app is bundled it includes that Network Extension target under _PlugIns_ directory in the main package:

<img width="215" height="192" alt="Image" src="https://github.com/user-attachments/assets/70e0af89-3ed5-4cf5-a9fc-cd54b3fee4ac" />

<hr/>

And that compiled .appex Network Extension target has the following structure inside of it:
<img width="281" height="190" alt="Image" src="https://github.com/user-attachments/assets/73fb10bc-b505-4605-9db7-9260fc8e9865" />

<hr/>

Now the question is, would including Network Extension and calling/using it be possible from Nucleus MacOS JVM application? 

Calling would possibly work by using **Kotlin/Native** to create a **dylib** and calling it from **JVM** side using **JNI/JNA** so the only questions I have right now would be about signing, bundling and shipping the .appex together with the app itself.

Under [MacOS packaging for App Store](https://nucleusframework.dev/en/docs/packaging/macos/#package-for-the-mac-app-store) section in Nucleus documentation it is shown how to set provisioning profiles and entitlements but I assume those are used for all embedded dylibs (etc.), what if we needed a different entitlements file or provisioning profile for one of the dylibs? 

Any ideas?

kdroidFilter:
@tomasjablonskis implemented as first-class DSL in draft PR #396. Nucleus copies the `.appex` into `Contents/PlugIns/`, signs it with its own entitlements + provisioning profile, and seals the outer app without `--deep`. JVM/jpackage, DMG/PKG and GraalVM native paths all covered.

```kotlin
macOS {
    entitlementsFile.set(file("packaging/app.entitlements"))
    appExtensions {
        extension("MyExtension") {
            appex(file("build/appex/MyExtension.appex"))                  // you build the .appex
            entitlements(file("packaging/networkextension.entitlements"))
            provisioningProfile(file("packaging/MyExtension.provisionprofile")) // → embedded.provisionprofile
        }
    }
}
```

[...]

PR #396, opened 2026-07-23
feat(packaging): embed & sign macOS app extensions (.appex)

kdroidFilter:
## Summary

Adds first-class support for embedding & signing macOS **app extensions** (`.appex`) — e.g. a Network Extension — inside a packaged app, addressing #394.

New DSL:

```kotlin
macOS {
    entitlementsFile.set(file("packaging/app.entitlements"))     // host-app entitlements
    appExtensions {
        extension("NetworkFilter") {
            appex(file("build/appex/NetworkFilter.appex"))                          // prebuilt .appex
            entitlements(file("packaging/extension/NetworkExtension.entitlements"))  // its OWN
            // provisioningProfile(file("packaging/NetworkFilter.provisionprofile"))
        }
    }
}
```

The plugin copies each `.appex` into `Contents/PlugIns/`, embeds its own `embedded.provisionprofile`, signs it inside-out with its own entitlements, then seals the outer app **without `--deep`** so the nested signature is preserved.

### What changed
- **DSL**: `MacAppExtensionSettings` / `MacAppExtension` + `macOS { appExtensions { } }`.
- **JVM/jpackage** (`AbstractJPackageTask`): embed + sign extensions before the final seal.
- **DMG/PKG** (`AbstractElectronBuilderPackageTask`): re-sign extensions in `resignApp` and in the ad-hoc `--deep` fallback (arm64) so installers preserve their entitlements.
- **GraalVM native** (`configureGraalvmApplication`): embed + ad-hoc sign extensions into the native `.app`, plus wire them into the GraalVM electron-builder task.
- **Example**: `examples/macos-appex-demo` (minimal `NEFilterDataProvider` compiled to a universal `.appex`, Compose UI that inspects its own `Contents/PlugIns`).

Nucleus does **not** build the `.appex` (that stays Xcode / Kotlin/Native); it copies + signs + seals a prebuilt one.

### No regressions by design
All new paths are guarded by `isEmpty()`/`isNotEmpty()` — behavior is byte-identical when `appExtensions` is unused.

## Test plan
- [x] Plugin compiles; `detekt` + `ktlint` pass
- [x] JVM `createDistributable` → `.appex` in `Contents/PlugIns`, `codesign --verify --deep --strict` OK, distinct entitlements (app `allow-jit`=1 / appex `allow-jit`=0, appex `networkextension`=1)
- [x] Self-signed named identity → app + appex carry the `Authority` (not adhoc)
- [x] GraalVM `packageGraalvmNative` (real native arm64 binary) → `.appex` embedded, `--deep --strict` OK
- [x] Regression: no-extension example (`extra-launcher-demo`) still packages, no `PlugIns`, codesign OK
- [ ] Verified by @tomasjablonskis against a real Network Extension via `publishDevToMavenLocal`

## Limitations
- GraalVM native images are always ad-hoc signed (embedded extension is ad-hoc too); configure `signing {}` for a notarized GraalVM DMG.
- *Activating* the extension at runtime uses the NetworkExtension management APIs (`NEFilterManager` / `NETunnelProviderManager`) via a native bridge — out of scope here; this is about signing/bundling/shipping the `.appex`.


tomasjablonskis:
@kdroidFilter verified that this works in https://github.com/NucleusFramework/Nucleus/issues/394#issuecomment-5253213110, tho there are few caveats (read more in that comment).

PR #436, opened 2026-08-04
feat(notification): implement platform notifications DSL

### LINKED ISSUE #397: [Feature] Add notification urgency to the common API; expose per-OS hints (transient, resident, expire-timeout) in the platform modules
## ⚠️ Is your feature request related to a problem? Please describe

The common notification API lets me set title, body, images, action buttons, and lifecycle
callbacks — but not the hints that control a notification's prominence and lifecycle:
urgency/priority, expire timeout, transient (shown but not kept in history), and resident
(kept after an action is invoked). There's no supported way to set these, even though the
backends already model them internally (e.g. the Linux backend has a `NotificationHints`
type carrying urgency/transient/resident/category that isn't surfaced anywhere).

## 💡 Describe the solution you'd like

- Add `urgency`/`priority` (e.g. `LOW`/`NORMAL`/`CRITICAL`) to the common API — it maps to
  a native equivalent on every known platform, so it fits the shared subset.
- Expose the remaining, platform-specific hints (`transient`, `resident`, `category`,
  `expire_timeout`) through the per-OS modules, honored where the OS supports them.

| Hint | Linux | Windows | macOS |
|---|---|---|---|
| Urgency / priority | ✅ | ✅ | ✅ |
| Expire timeout | ✅ | ⚠️ partial | ❌ |
| Transient (no history) | ✅ | ❌ | ❌ |
| Resident (survive an action) | ✅ | ❌ | ❌ |
| Category | ✅ | ⚠️ partial | ⚠️ partial |

## 🤚 Do you want to develop this feature yourself?

- [x] Yes
- [ ] No

rocketraman:
## 🚀 Description

Adds first-class **per-platform option blocks** to the cross-platform notification DSL. The common `notification { }` builder now accepts `linux { }`, `macos { }`, and `windows { }` blocks, each applied only when the notification is delivered on that OS:

```kotlin
notification(title = "Build failed", message = "see logs") {
    button("Open") { openLogs() }

    // Behavior that differs per OS lives where the semantics are unambiguous:
    linux   { urgency = Urgency.CRITICAL; transient = true; category = "device.error" }
    macos   { interruptionLevel = InterruptionLevel.TIME_SENSITIVE; relevanceScore = 0.9 }
    windows { scenario = ToastScenario.URGENT; duration = ToastDuration.LONG }
}
```

The common fields (title, message, image, buttons, callbacks) remain the shared, genuinely-uniform base; anything whose behavior diverges between platforms is set in the matching block and is a no-op elsewhere.

Highlights:
- New `LinuxNotificationScope` / `MacNotificationScope` / `WindowsNotificationScope`, typed with the existing platform enums.
- `notification-common`'s platform dependencies are now `api` (were `implementation`) so consumers can name `Urgency` / `InterruptionLevel` / `ToastScenario` directly in the blocks.
- `notification-windows` gains `ToastScenario.URGENT`, a `ToastDuration` enum, and `duration` toast support (used by the `windows { }` block).

Resolves #397

## 📄 Motivation and Context

#397 asked for notification urgency and per-OS hints (transient, resident, category, expire-timeout) in the API. The first cut mapped a common `urgency`/`priority` enum onto each platform, but that abstraction is dishonest: `CRITICAL` breaks through Do Not Disturb on Linux and Windows, whereas on macOS that requires an Apple entitlement most apps don't have, so it can only reach `timeSensitive` — which merely re-prioritizes. There is no uniform 3-level ladder to map to (Windows toasts don't even have a graded "low").

Following the discussion on the issue, this PR takes the opposite approach: **don't fake a shared abstraction for behavior that isn't shared.** The common surface stays limited to the truly-portable structural fields, and platform-specific behavior is expressed through explicit per-platform blocks the caller opts into. This is also the ergonomic answer for a single-codebase multiplatform desktop app — the OS is a runtime dimension of one JVM artifact (not a compile target, so `expect/actual` can't express it), and the blocks let you configure all three platforms in one call site with full fidelity instead of hand-rolling `when (Platform.Current)`.

Environment: Fedora Linux, KDE Plasma 6 (Wayland), JDK 21 (Temurin).

- **Unit tests** — `NotificationDslTest` (blocks populate only their own scope, buttons still work, repeated blocks merge) and `ToastXmlBuilderTest` (`scenario="urgent"` and `duration="short"/"long"` emitted correctly, omitted by default).
- **Static checks** — `ktlintCheck` and `detekt` pass on the changed modules.
- **Linux end-to-end, on the wire** — ran the real common API → `LinuxDispatcher` → JNI → D-Bus against a mock `org.freedesktop.Notifications` server on an isolated `dbus-run-session` bus, and asserted the captured hints: `linux { urgency = CRITICAL }` → `urgency=2`, `LOW` → `0`, and `transient/resident/category="im.received"/expireTimeout=3000` all land correctly. A no-block notification emits identical output to the previous API (empty hints, `expire_timeout=-1`), confirming backward compatibility.
- **macOS / Windows** — verified at compile + unit level (no macOS/Windows hosts available); the `windows { }` path is covered by the XML builder test, and the `macos { }` mapping sets `interruptionLevel`/`relevanceScore`/`subtitle` on `NotificationContent`, which is already wired to native.
- `examples/nucleus-demo` compiles with a new "Platform-Specific Options" section demonstrating the blocks.

Backward-compatibility note: this is source-compatible — the `notification(...)` parameters are unchanged, `button()` still works, the trailing-lambda receiver was renamed (`NotificationButtonBuilder` → `NotificationBuilder`, with a `@Deprecated typealias` kept), and the trailing-lambda parameter was renamed (`buttons` → `block`); no in-repo caller referenced either name. It is not binary-compatible (the lambda receiver type changed), so downstream code should be recompiled.

## 📦 Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)

## ✅ Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,7 @@ Each module is published independently to Maven Central — use them together or
 | `nucleus.decorated-window-jewel` | Jewel (IntelliJ theme) integration |
 | `nucleus.decorated-window-material2` | Material 2 integration |
 | `nucleus.decorated-window-material3` | Material 3 integration |
+| `nucleus.notification-common` | Cross-platform notification DSL with per-platform (`linux`/`macos`/`windows`) option blocks |
 | `nucleus.notification-macos` | macOS User Notifications |
 | `nucleus.notification-windows` | Windows Toast Notifications |
 | `nucleus.notification-linux` | Freedesktop Desktop Notifications |
--- a/examples/nucleus-demo/src/main/kotlin/com/example/demo/CommonNotificationsScreen.kt
+++ b/examples/nucleus-demo/src/main/kotlin/com/example/demo/CommonNotificationsScreen.kt
@@ -31,10 +31,13 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.unit.dp
+import dev.nucleusframework.notification.InterruptionLevel
 import dev.nucleusframework.notification.common.NotificationHandle
 import dev.nucleusframework.notification.common.NotificationManager
 import dev.nucleusframework.notification.common.NotificationResult
 import dev.nucleusframework.notification.common.notification
+import dev.nucleusframework.notification.linux.Urgency
+import dev.nucleusframework.notification.windows.ToastScenario
 
 private const val EVENT_LOG_MAX = 20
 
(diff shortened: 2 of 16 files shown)


kdroidFilter:
Thanks :)

Issue #52, opened 2026-02-23
Root key feature


StefanOltmann:
Hydraulic Conveyor offers a very convenient feature: you provide a single root key, typically as a passphrase, and it deterministically derives platform-specific self-signing keys from it. This saves a significant amount of setup and key management effort.

On macOS in particular, this is important because enrolling in the paid Apple Developer Program is not strictly required just to sign your own apps. You can sign macOS applications with your own certificate. The crucial part is that the app is properly signed, not that the certificate was issued by Apple. That often comes as a surprise.

More details about this approach are available here:
https://conveyor.hydraulic.dev/21.1/configs/keys-and-certificates/#root-key

It would be highly beneficial if Nucleus could support a similar model: accept a single root key via environment variables, for example through GitHub Secrets, and automatically derive self-signed binaries for Windows, macOS, and Linux from it.

PR #445, opened 2026-08-05
fix(tao): wayland window resize content jumping fix

### LINKED ISSUE #444: Tao wayland window resize: window content jumps and stretches on Fedora
## 🐛 Describe the bug

On native Wayland, resizing a `NucleusBackend.Tao` window by dragging an edge
makes the window contents visibly jump and stretch instead of tracking the
pointer. It is worse the larger the window, and it does not happen with other
applications on the same desktop (GTK3 and Qt apps both resize cleanly).

Investigation found three independent defects stacked on top of each other, all
specific to the Tao backend's split-surface architecture (GTK owns the toplevel;
Compose renders into a child `wl_subsurface`):

1. **The content lags the window frame by ~32 ms.** GDK acks a `configure` and
   commits the matching `xdg_surface.set_window_geometry` about 7 ms later, on
   its own frame clock, while our buffer for that size only exists at ~37 ms.
   Because the compositor pins the subsurface at its origin and leaves the rest
   uncovered, the mismatch is a *positional* error: the content detaches from
   the dragged edge and snaps back. Measured: the two disagree for ~95 % of a
   drag, by 27 px p50 / 35 px p90.

2. **Skia is told a size the GL framebuffer does not have.**
   `wl_egl_window_resize` only records a *pending* size — the buffer behind the
   default framebuffer is not reallocated until the next `eglSwapBuffers`. The
   render target wraps that framebuffer (`fbId = 0`) but was built from the size
   we had just requested. With `SurfaceOrigin.BOTTOM_LEFT`, overstating the
   height by N puts the whole frame N rows off the top of the real drawable,
   leaving an N-pixel band of clear colour at the top of the window. During a
   drag that band appears and disappears on ~40 % of frames, sized by drag
   speed. This is the flicker that is actually visible.

3. **A residual stretch.** Once (2) is fixed by rendering at the framebuffer's
   real size, that size is still one resize step behind, so presenting it
   against a viewport destination set to the current window size scales it by
   that step (~2 % on a fast drag).

## ⚠️ Current behavior

During an edge drag the window contents jump by roughly the height of the title
bar on a large fraction of frames, and the content detaches from the dragged
edge and snaps back. The effect scales with drag speed and window size.

## ✅ Expected behavior

Window contents stay locked to the window while it is being resized, as they do
in native GTK3 and Qt applications on the same compositor.

## 💣 Steps to reproduce

1. Run any Tao-backed sample on a native Wayland session, e.g.
   `./gradlew :examples:tao-demo:run`.
2. Enlarge the window to a substantial fraction of the screen (the artifact
   scales with window area — it is easy to miss on a small window).
3. Grab the bottom-right corner and drag it back and forth reasonably quickly.
4. Watch the window chrome (title bar / toolbar row) rather than the center.

## 📷 Screenshots

https://github.com/user-attachments/assets/30622812-a0d2-4ee1-84a1-1049273c29d1

Frame-by-frame analysis of a capture of one drag: the window's own top-left stays
at a fixed position in every frame, while the *content* is displaced on ~41 % of
frames and back on the next one. The displacement is exactly the size of that
frame's resize increment — confirmed on 11 of 11 frames where the window grew,
across increments from 8 px to 60 px — so it scales with drag speed rather than
being a fixed offset. It appears as a band of clear colour along the top edge,
which is the signature of defect (2) above: the `SurfaceOrigin.BOTTOM_LEFT`
y-flip applied against a framebuffer that is one resize step smaller than Skia
believes.

## 📱 Tech info

 - Device: AMD Navi 48 (radv)
 - OS: Fedora 44, KWin 6.7.3, native Wayland session, 3 × 4K displays at
   fractional scale 1.3
 - Library/App version: `decorated-window-tao` @ `main`, Compose 1.11.1,
   Temurin 21

Notes:
 - Wayland only. The X11/XWayland path is unaffected.
 - Not reproducible under `kwin_wayland --virtual`: the headless compositor has
   no real vblank, which changes the frame timing the defects depend on.

rocketraman:
## 🚀 Description

Fixes the Tao/Wayland resize artifact in three steps, one commit each.

**1. Size the Compose subsurface with `wp_viewport`, not its buffer.**
Bind `wp_viewporter` and drive `wp_viewport.set_destination` from the newest
`configure`, instead of letting the surface size follow whatever buffer we last
rendered. The surface then tracks the window immediately. `wp_viewporter` is an
extension, so its `wl_interface` tables are not exported by libwayland-client
and cannot be `dlsym`-ed like the core ones — they are hand-authored against the
layouts already declared in `nucleus_tao_egl.c`.

**2. Render at the EGL surface's real size, not the requested one.**
Query the surface with `eglQuerySurface` and build the render target, Skia
surface and scene size from that. Skia and GL then always agree, so the
`SurfaceOrigin.BOTTOM_LEFT` y-flip is correct and nothing is displaced.

The sanity bound on the queried size must be **symmetric** — the framebuffer
lags in both directions (smaller while growing, larger while shrinking), and an
asymmetric bound silently disables the fix for shrinking.

**3. Over-allocate the EGL buffer and crop it with `wp_viewport.set_source`.**
Allocate the `wl_egl_window` on a 128 px grid *plus a whole grid of headroom*,
so the allocation does not change while dragging. Every frame is then rendered
at the true window size and the window rect is cropped back out, presented 1:1.
The initial allocation in `nativeAttachWayland` is bucketed too — creating at
the exact size left the *first* drag with no headroom, which is where the error
is most visible, since the scale error at a boundary crossing is one frame of
drag motion over the window size (~3 % at 1024 px, ~1 % at 3800 px).

The source rectangle is clamped to the buffer actually in flight: a source rect
running past the buffer is a fatal protocol error. A source under half the
destination is discarded rather than cropped — before the first `configure` is
applied the surface can read back as 1×1, and stretching that across the window
would flash.

Degrades cleanly on compositors without `wp_viewporter` (falls back to
buffer-derived sizing) and is a no-op on X11.

## 📄 Motivation and Context

Fixes #444 .

Each commit addresses a distinct defect; they are ordered as discovered and each
builds on the previous one. Notably, (2) is invisible to every surface-geometry
metric — the surface size is correct throughout, and the error lives entirely
inside the rendered buffer. It was found by frame-by-frame analysis of a screen
capture, after protocol tracing repeatedly failed to see it.

Environment: Fedora 44, KWin 6.7.3 native Wayland, 3 × 4K at fractional scale
1.3, AMD Navi 48 / radv, Temurin 21, Compose 1.11.1.

Measured on the reporting machine with `WAYLAND_DEBUG` traces of a resize drag
on a 2842×1840 window:

| | before | after (1) |
| --- | --- | --- |
| time content size ≠ window geometry | 94.6 % | ~26 % |
| \|mismatch\| p50 | 27 px | 0 px |
| \|mismatch\| p90 | 35 px | 9 px |
| share of drag exactly correct | 5 % | 74 % |

For (2), a frame-content probe over a resize drag — frames whose top band
collapsed to under 20 px: **108 of 257 → 0**; over a combined grow+shrink drag,
**1 of 434**.

For (3), counting viewport commits where source == destination (the direct
precondition for no scaling), over a first drag from the default window plus an
aggressive grow/shrink drag: **897 of 897 presented exactly 1:1, 0 scaled, 0
protocol errors**.

Also verified: rendering is unchanged and sharp at rest (an incorrect
destination here would blur the window permanently), and the window renders
correctly after dragging.

**Known residual:** a small amount of stretch is still visible on the *initial*
drag. Everything after that is clean. The remaining case is a drag that gains a
full grid step (128 px) inside one frame, where the window outruns even the
headroom; raising `EGL_ALLOC_GRID` to 256 would double the margin at the cost of
a few MB of buffer. Left as-is pending review feedback.

**Testing gaps:** (3) was verified headless plus by direct observation on the
reporting rig; the wire-level counts come from `kwin_wayland --virtual`, which
has no real vblank. The artifact itself is not reproducible there, so those
counts confirm the mechanism, not the perceived result.

## 📦 Types of changes

## ✅ Checklist

### DIFF
--- a/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
+++ b/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
@@ -147,6 +147,21 @@ internal object NativeTaoEglBridge {
         commitNow: Boolean,
     )
 
+    /**
+     * The EGL window surface's **actual** size, packed `(width shl 32) or height`,
+     * or 0 if unavailable. Wayland only in practice.
+     *
+     * Deliberately not the size last passed to [nativeResize]:
+     * `wl_egl_window_resize` records a *pending* size and the buffer behind the
+     * GL default framebuffer is only reallocated on the next `eglSwapBuffers`.
+     * Between those two points this reports the OLD size, while our own state
+     * has already moved to the new one — and anything that wraps `fbId = 0`
+     * (i.e. `BackendRenderTarget.makeGL`) is wrapping a framebuffer of that old
+     * size. See `docs/linux-wayland-resize-latency.md`.
+     */
+    @JvmStatic
+    external fun nativeSurfaceSize(handle: Long): Long
+
     /** Pumps the back-buffer to screen via `eglSwapBuffers`. */
     @JvmStatic
     external fun nativePresent(handle: Long)
--- a/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
+++ b/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
@@ -136,14 +136,23 @@ internal object NativeTaoEglBridge {
      * safe from the event-loop thread **while no swap is in flight** — the swap
      * thread commits the same surface. Callers must hold that gate themselves.
      *
-     * No-op on X11, on compositors without `wp_viewporter`, and when the
-     * destination is unchanged and [commitNow] is false.
+     * [srcW]/[srcH] crop the window rect out of the buffer, which is
+     * over-allocated on a coarse grid so a drag rarely reallocates it. **They
+     * must not exceed the buffer currently in flight** — a source rectangle
+     * running past the buffer is a fatal `wp_viewport` protocol error — so the
+     * caller clamps them to the size reported by [nativeSurfaceSize]. Pass 0 to
+     * fall back to the destination size.
+     *
+     * No-op on X11, on compositors without `wp_viewporter`, and when nothing
+     * changed and [commitNow] is false.
      */
     @JvmStatic
     external fun nativeSetViewportDestination(
         handle: Long,
         logicalW: Int,
         logicalH: Int,
+        srcW: Int,
+        srcH: Int,
         commitNow: Boolean,
     )
 
--- a/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
+++ b/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
@@ -121,6 +121,32 @@ internal object NativeTaoEglBridge {
         yLogical: Int,
     )
 
+    /**
+     * Wayland only: sets the content surface's `wp_viewport` destination —
+     * the size the surface occupies in surface (logical) units, independent of
+     * the buffer's size. The compositor scales the buffer to fit.
+     *
+     * This decouples "how big the window is" from "how big the buffer we last
+     * finished rendering is". Those are ~30 ms apart during a resize drag, and
+     * that gap is the resize artifact (see
+     * `docs/linux-wayland-resize-latency.md`).
+     *
+     * [commitNow] makes the change land immediately instead of waiting for the
+     * next `eglSwapBuffers`. It issues a `wl_surface.commit`, so it is only
+     * safe from the event-loop thread **while no swap is in flight** — the swap
+     * thread commits the same surface. Callers must hold that gate themselves.
+     *
+     * No-op on X11, on compositors without `wp_viewporter`, and when the
+     * destination is unchanged and [commitNow] is false.
+     */
+    @JvmStatic
+    external fun nativeSetViewportDestination(
+        handle: Long,
+        logicalW: Int,
+        logicalH: Int,
+        commitNow: Boolean,
+    )
+
     /** Pumps the back-buffer to screen via `eglSwapBuffers`. */
     @JvmStatic
     external fun nativePresent(handle: Long)
--- a/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/scene/TaoComposeSceneHostLinux.kt
+++ b/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/scene/TaoComposeSceneHostLinux.kt
@@ -222,6 +222,10 @@ internal class TaoComposeSceneHostLinux(
      */
     private var swapThread: SwapThread? = null
 
+    /** Real framebuffer size the cached Skia surface was built for. */
+    private var lastFbWidthPx: Int = 0
+    private var lastFbHeightPx: Int = 0
+
     private var widthPx: Int = 0
     private var heightPx: Int = 0
     private var scale: Float = 1f
@@ -1008,6 +1012,33 @@ internal class TaoComposeSceneHostLinux(
     }
 
     /**
+     * Size to render at, from the EGL surface size packed by
+     * `nativeSurfaceSize`.
+     *
+     * The framebuffer lags the requested size in **either** direction: smaller
+     * while the window grows, larger while it shrinks. An earlier version of
+     * this guard required `w <= widthPx`, which silently disabled the whole fix
+     * for shrinking — and matched the report exactly ("almost gone when making
+     * the window larger, still very visible when making it smaller"). The bound
+     * is therefore symmetric.
+     *
+     * Its only job is to reject a *degenerate* report: before the first
+     * configure is applied the EGL surface can return something tiny, and laying
+     * the scene out into that leaves the window unmapped. A real resize step is
+     * a handful of pixels, so a factor-of-two window either way never rejects a
+     * legitimate value.
+     */
+    private fun resolveFramebufferSize(packed: Long): IntSize {
+        if (packed == 0L) return IntSize(widthPx, heightPx)
+        val w = (packed ushr 32).toInt()
+        val h = (packed and 0xFFFFFFFFL).toInt()
+        val sane =
+            w in (widthPx / 2)..(widthPx * 2) &&
+                h in (heightPx / 2)..(heightPx * 2)
+        return if (sane) IntSize(w, h) else IntSize(widthPx, heightPx)
+    }
+
+    /**
      * Pushes the *current* window size to the content surface's `wp_viewport`
      * destination, so the surface occupies the window rect regardless of how
      * stale the buffer we last rendered is.
@@ -1188,12 +1219,44 @@ internal class TaoComposeSceneHostLinux(
         // is current — applyPendingNativeResize closes the stale Skia cache.
         applyPendingNativeResize()
 
+        // Render at the framebuffer's REAL size, not the size we requested.
+        //
+        // `wl_egl_window_resize` only records a pending size; the buffer behind
+        // the GL default framebuffer is not reallocated until the next
+        // `eglSwapBuffers`. `BackendRenderTarget.makeGL(fbId = 0)` wraps that
+        // same framebuffer, so passing the requested size tells Skia the drawable
+        // is taller than it is — and with `SurfaceOrigin.BOTTOM_LEFT` Skia maps
+        // y=0 to GL row `height - 1`, off the top of the real drawable. The whole
+        // frame is displaced by the difference, leaving a band of clear colour at
+        // the top of the window, `delta-height` tall, flickering for the whole
+        // drag. Measured: the framebuffer is exactly one resize step behind on
+        // every resizing frame. See docs/linux-wayland-resize-latency.md.
+        //
+        // Drawing smaller than the window is safe: the `wp_viewport` destination
+        // already maps whatever we render onto the current window rect.
+        val packedFb = NativeTaoEglBridge.nativeSurfaceSize(attachmentHandle)
+        val fbSize = resolveFramebufferSize(packedFb)
+        val fbW = fbSize.width
+        val fbH = fbSize.height
+        if (fbW != lastFbWidthPx || fbH != lastFbHeightPx) {
+            cachedSurface?.close()
+            cachedSurface = null
+            cachedRt?.close()
+            cachedRt = null
+            lastFbWidthPx = fbW
+            lastFbHeightPx = fbH
+        }
+        if (scene?.size != IntSize(fbW, fbH)) {
+            scene?.size = IntSize(fbW, fbH)
+            updateWindowInfoSize()
+        }
+
         var surface = cachedSurface
         if (surface == null) {
             val rt =
                 BackendRenderTarget.makeGL(
-                    width = widthPx,
-                    height = heightPx,
+                    width = fbW,
+                    height = fbH,
                     sampleCnt = 0,
                     stencilBits = 8,
                     fbId = 0,
--- a/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/scene/TaoComposeSceneHostLinux.kt
+++ b/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/scene/TaoComposeSceneHostLinux.kt
@@ -591,6 +591,9 @@ internal class TaoComposeSceneHostLinux(
         // The redraw gate may have latched while hidden (invalidations with no
         // draw ever arriving); clear it so the re-arm below goes through.
         redrawPending.set(false)
+        // The swap in flight during the last configure has finished by now, so a
+        // destination whose commit had to be skipped can land here.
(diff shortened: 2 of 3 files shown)


kdroidFilter:
Thank you so much, it looks great, I'll reread it.

kdroidFilter:
I tested the three commits on GNOME/Mutter (Wayland session) before reviewing the code, and the results don't line up with the premise:

| build | result |
| --- | --- |
| unpatched `main` | no resize artifact at all |
| commits 1+2 | an artifact appears, resembling the reported symptom |
| commits 1+2+3 | much worse — during a drag the top band of the window is visibly duplicated partway down, which reads as a hard flash |

So on this setup the patch is a regression, and the defect it targets doesn't reproduce. That doesn't invalidate your measurements — it suggests the root cause is specific to your rig (KWin 6.7.3, 3×4K at fractional scale 1.3, radv) rather than general to Tao/Wayland, and that the fix can't be unconditional.

A hypothesis for the direction of the regression, which I haven't confirmed with a trace: commit 2 sizes Skia from the *queried* framebuffer, which lags a resize step behind, so on its own it under-fills the window — it only works paired with commit 1's viewport stretching the result back onto the window rect. Where the buffer/geometry gap doesn't manifest, that pair turns "no artifact" into "one scaled frame per resize step". Commit 3's 128 px grid then turns that step into a 128–256 px displacement, which would match the duplication I see.

Two code issues that stand independently of any of the above, both in commit 3:

1. **Popups.** `TaoPopupSceneLayerLinux` shares `nativeAttachWayland`/`nativeResize`, so it gets bucketed buffers, but it never drives the viewport — no destination is ever set, so its surface size stays `buffer_px / buffer_scale` = the bucketed size. A 200×100 tooltip becomes a 384×256 surface, and its render target is built at 200×100, so the margin is uninitialised. Gating the grid on a destination having been set (e.g. on `viewport_w`) fixes it and makes "bucketed" and "cropped" impossible to get out of step.
2. **`resolveFramebufferSize`'s upper bound.** The grid overshoots by an absolute step, not a ratio — a 100 px axis allocates 256 px — so `w in (widthPx / 2)..(widthPx * 2)` rejects a legitimate size on any window under 128 px on either axis and falls back to `widthPx`/`heightPx`, handing Skia a size the framebuffer doesn't have. That's the same `BOTTOM_LEFT` displacement commit 2 exists to prevent. The lower bound is the one doing useful work.

Also, `docs/linux-wayland-resize-latency.md` is referenced 5 times across the three files but isn't in the PR — it sounds like the most useful part of this investigation.

Where I'd go next: figure out which part of your setup is load-bearing — the ~30 ms render latency, GDK's ack timing under KWin, or the fractional scale. If you can share a `WAYLAND_DEBUG` capture of a drag on your machine I'll capture the same on Mutter and diff them.

kdroidFilter:
Went looking for prior art on this exact class, and SDL has already been through all of it.

**[libsdl-org/SDL#4563](https://github.com/libsdl-org/SDL/issues/4563)** — *"Buffers with old size can get commited already after ack_configure on window resize"*. This is defect (1) from #444 in SDL's own words, with a Mutter trace:

> the compositor sends `xdg_toplevel::configure(800, 600)` which gets acked and the new surface state gets commited right away — but the new state still contains the old 1280x720 buffer, as the new one gets created only after that commit.

Their fix direction is in the commit they reference, [`785dfdfbe`](https://github.com/libsdl-org/SDL/commit/785dfdfbe2633c203c1fe74445c98019cb8052dc): **"Ack configure after the buffer size has been actually changed."** Same answer as `flutter_linux`'s `wait_for_frame()`.

**[libsdl-org/SDL#4326](https://github.com/libsdl-org/SDL/issues/4326)** is the opposing constraint, and it carries the line that matters here — SDL's Wayland GLES path used to be commented:

```c
// Wayland-EGL forbids drawing calls in-between SwapBuffers and wl_egl_window_resize
```

**[NVIDIA/egl-wayland#53](https://github.com/NVIDIA/egl-wayland/pull/53)** (merged) and its issue **[#52](https://github.com/NVIDIA/egl-wayland/issues/52)** cover calling `wl_egl_window_resize` off the thread the EGLSurface is current on, confirmed in-thread by an NVIDIA engineer. Not our bug — `applyPendingNativeResize` runs after `nativeMakeCurrent`, so we resize with the context current on the calling thread — but worth keeping as a documented invariant, because moving the resize onto the swap thread would have to preserve it.

**Where this lands on commit 2.** Our per-frame order is `makeCurrent` → `applyPendingNativeResize` (`wl_egl_window_resize`) → draw → `releaseCurrent` → *[swap thread]* `eglSwapBuffers`. So the resize is already the first thing after the previous swap with no draws in between — the SDL constraint is satisfied. But Mesa reallocates lazily, at back-buffer acquisition, i.e. **during the draw** — and commit 2 calls `eglQuerySurface` *before* that draw. It gets the pre-realloc size, sizes Skia from it, and the draw then goes into the newly-allocated buffer. With `SurfaceOrigin.BOTTOM_LEFT` that displaces the whole frame by the difference.

Which inverts the premise: the framebuffer size at draw time is *deterministic* — it's what we last passed to `wl_egl_window_resize` — and querying it is what introduces the error. `main` sizing Skia from `widthPx`/`heightPx` is the correct value. Commit 3's 128 px grid then turns that difference from a resize step into a full grid step, which is what I see as content duplication on Mutter.

**Suggested direction.** The root-cause fix is acking after the buffer exists, and GDK owns the ack here, on its own frame clock. So rather than taking the ack over, the lever is to render *before* GDK's paint phase runs — driving the frame from the GDK frame clock (`update`/`before-paint`) instead of tao's separate draw tick, so our buffer and GTK's geometry commit land in the same cycle. That keeps the swap thread exactly as it is; it only changes what triggers a frame, not what presents it.

I can't validate any of this locally — the defect doesn't reproduce on my setup, unpatched `main` is clean here — so this is a mechanism, not a tested patch. But it's cheap to check against your rig: if the displacement really is the lazy realloc, then dropping commit 2's query and keeping `widthPx`/`heightPx` should not reintroduce the band you measured.

rocketraman:
@kdroidFilter

> But it's cheap to check against your rig: if the displacement really is the lazy realloc, then dropping commit 2's query and keeping widthPx/heightPx should not reintroduce the band you measured.

Ran it. Instrumented eglQuerySurface before the draw and again after flushAndSubmit, context still current.

The buffer never changes size across the draw — pre == post on 726/726 frames (KWin 6.7.3, Mesa/radv). So Mesa isn't reallocating at back-buffer acquisition here, the pre-draw query isn't stale, and it isn't what displaces the frame. With the query in place Skia matched the real buffer on 360/360 frames.

Dropping the query doesn't remove the band, it makes one permanent: 366/366 frames mismatched, mean 213 px, visible with nothing being dragged.

But that's dropping commit 2 while keeping commit 3, which isn't your proposal. Distinct EGL buffer sizes over one resize sweep: base 62, commit 1 only 61 — both exactly equal to the configure sizes — all three commits: 8 (the 128 px grid). So without commit 3 the buffer is the window size and commit 2 is a no-op. You're right that 2 and 3 are one change, not two; 3 creates the discrepancy 2 corrects. That's on me to justify as a pair or drop both.

On commit 1 alone: it takes time-content-≠-geometry from 78.8% to 0.8% in a headless capture — and the artifact is still there on my rig. So the surface-geometry metric this PR optimises isn't the defect. Chasing that now; more when I have it.

> Where I'd go next: figure out which part of your setup is load-bearing — the ~30 ms render latency, GDK's ack timing under KWin, or the fractional scale. If you can share a WAYLAND_DEBUG capture of a drag on your machine I'll capture the same on Mutter and diff them.

Here is the trace, about 10 seconds of quickly dragging the bottom right corner of the tao-demo : [trace-base.log.gz](https://github.com/user-attachments/files/30755097/trace-base.log.gz)

I note also another user appears to have the same issue on MacOS which would totally invalidate the fix direction taken so far here: https://kotlinlang.slack.com/archives/C0BJ0GTE2/p1785777991758679?thread_ts=1785755862.604249&cid=C0BJ0GTE2. Have asked that user to provide more information.

rocketraman:
@kdroidFilter Two follow-ups, plus a correction to what I said above.

## Both code issues confirmed

1. **Popups.** `nativeSetViewportDestination` is called only from `TaoComposeSceneHostLinux`; `TaoPopupSceneLayerLinux` and `TaoStandalonePopupHostLinux` reach `egl_bucket` through `nativeAttachWayland`/`nativeResize` and never set a destination. Gating the grid on `viewport_w` is the right fix.
2. **`resolveFramebufferSize`.** Confirmed — the grid overshoots by an absolute step, so the upper bound rejects a legitimate size on any window under 128 px per axis. Only the lower bound does useful work.
3. `docs/linux-wayland-resize-latency.md` references should be removed — it's referenced from `main` too via #447, so if we find a way to move forward on this this will get fixed.

These will all get fixed but a lot will depend on where this PR ends up.

## Correction on commits 2 and 3

I said above they were "one change, not two, to justify as a pair or drop both". The framing was wrong. Commit 1 doesn't leave the buffer alone — it hands the compositor a stale buffer to stretch onto the window rect, and that stretch is observable, which is what 2 and 3 were written to remove. They're the back half of commit 1, not additions to it. Combined with commit 1 alone still showing the artifact, the three are a single change; there's no subset to drop.

## Rendering in step with the resize doesn't fix it

Testing your frame-clock direction in its simplest form — lay out and render inside the resize event, `scene.size` throttle removed:

| | frames | displaced | mean | p50 | p90 | max |
| --- | --- | --- | --- | --- | --- | --- |
| current (deferred) | 669 | 23% | 49.5 px | 52 | 78 | 180 |
| synchronous frame | 1402 | 70% | 13.2 px | 10 | 28 | 66 |

("displaced" = the frame's layout size vs the window size at the moment it reaches the compositor.)

Twice the frames, every magnitude measure down 60–80%, artifact unchanged — and unusable during a drag, since the wait blocks the thread that dispatches input. Shortening the render→present window isn't sufficient; that's consistent with your read that the ack is the lever, but it also means the ack change has to be tested directly rather than approximated.

## Where this leaves it

Two independent metrics — surface-vs-geometry, and layout-vs-window-at-present — now improve substantially while the visible symptom doesn't move. Neither is measuring the actual defect.

The question I'd most like an answer to is still the one your Mutter results raise: the same change that measurably fixes this on KWin regresses it on Mutter. If the macOS report holds up, the cause sits above all three commits and this whole approach is aimed at the wrong layer. Let me know what you find out from the wayland trace given in the previous message -- or if you'd rather have me do the diff and analysis post your trace and I'll take a look.

rocketraman:
More updates:

Fractional scale is ruled out: I normally scale at 130% -- at 100% on this rig the artifact is just as bad. That leaves the ~30 ms render latency and GDK's ack timing under KWin, which the trace above should show against your Mutter capture.

One test-setup question before we chase a compositor difference: how large was your window? #444 notes the artifact scales with window area and is easy to miss on a small one — the measurements here are from 2842×1840. If your "unpatched `main` is clean" run was on a modest window, that alone could account for the divergence, and it's cheaper to rule out than anything else on the list.

kdroidFilter:
Can you check with https://github.com/NucleusFramework/Nucleus/pull/462 ?

rocketraman:
> Can you check with #462 ?

@kdroidFilter It does not solve the problem:

[Screencast_20260805_193511.webm](https://github.com/user-attachments/assets/ff92c2fc-6546-41e8-aee7-a25e957af87c)

kdroidFilter:
> > Can you check with #462 ?
> 
> @kdroidFilter It does not solve the problem:
> 
> [Screencast_20260805_193511.webm](https://github.com/user-attachments/assets/ff92c2fc-6546-41e8-aee7-a25e957af87c)
> 

Are you on x11 or wayland ?

rocketraman:
> Are you on x11 or wayland ?

Wayland

kdroidFilter:
> > Are you on x11 or wayland ?
> 
> Wayland

Can you check with x11?

kdroidFilter:
This PR causes huge regressions on Ubuntu GNOME

rocketraman:
> > > Are you on x11 or wayland ?
> > 
> > 
> > Wayland
> 
> Can you check with x11?

With xwayland, I have the same problem (app run with `env GDK_BACKEND=x11 ./gradlew :examples:tao-demo:run`). I don't have a native X environment to experiment on.

rocketraman:
> This PR causes huge regressions on Ubuntu GNOME

Yeah I don't think this PR is the right approach to fix this problem. But I've spent quite a few hours with Claude trying to fix this and haven't gotten anywhere with it.

kdroidFilter:
> > This PR causes huge regressions on Ubuntu GNOME
> 
> Yeah I don't think this PR is the right approach to fix this problem. But I've spent quite a few hours with Claude trying to fix this and haven't gotten anywhere with it.

I managed to reproduce the problem on a virtual machine with Fedora GNOME, but #462 fixed it, so it's strange. I'll try with KDE.

kdroidFilter:
@rocketraman can you recheck with #462  ?

rocketraman:
> @rocketraman can you recheck with #462 ?

Still have issues, though the problems seem a bit different now. Video here: https://github.com/NucleusFramework/Nucleus/issues/444#issuecomment-5205498829

rocketraman:
I'll close this since this fix didn't go anywhere useful.

PR #447, opened 2026-08-05
fix(tao/linux): declare the Compose subsurface's opaque region

### LINKED ISSUE #446: Tao wayland opaque region not set
## 🐛 Describe the bug

The Compose content subsurface on Wayland never calls
`wl_surface.set_opaque_region` — the opcode is not even defined in
`nucleus_tao_egl.c` — even though comments throughout the file describe the
surface as opaque.

Without a declared opaque region a compositor must assume the surface is
translucent, and therefore cannot cull anything drawn beneath it: GTK's toplevel,
and on Wayland the CSD drop shadow GTK now paints there. It alpha-blends the
whole stack across the entire window, every frame.

That cost lands on the compositor's frame timing, which is what GDK's frame
clock waits on, which in turn caps how often window geometry can be committed —
i.e. how smoothly the window edge tracks the pointer during a resize drag.

## ⚠️ Current behavior

The compositor blends layers it could discard. During a resize drag the window
edge advances in coarser, less frequent steps than it needs to.

Measured on a 2842×1840 window under KWin 6.7.3 during a drag:

| | value |
| --- | --- |
| toplevel frame callback latency | 16.9 ms |
| median gap between window-edge moves | 25.6 ms |

For reference on the same machine, a native GTK3 app moves its window edge every
18.1 ms and a Qt app every 16.0 ms.

## ✅ Expected behavior

The opaque region is declared so the compositor can skip what is hidden, and the
window edge tracks the pointer at a cadence comparable to native toolkits.

## 💣 Steps to reproduce

1. Run a Tao-backed sample on a native Wayland session with an opaque window,
   e.g. `./gradlew :examples:tao-demo:run`.
2. Capture a protocol trace: `WAYLAND_DEBUG=1 … > /tmp/wl.log 2>&1`.
3. `grep set_opaque_region /tmp/wl.log` — no matches.
4. Resize the window by dragging an edge and compare the smoothness of the edge
   against a native GTK3 or Qt application at a similar size.

## 📱 Tech info

 - Device: AMD Navi 48 (radv)
 - OS: Fedora 44, KWin 6.7.3, native Wayland session, 3 × 4K displays at
   fractional scale 1.3
 - Library/App version: `decorated-window-tao` @ `main`, Compose 1.11.1,
   Temurin 21

Note: the numbers above were measured *before* the CSD shadow rework
(`de5e20ff`), i.e. with one more full-window layer in the stack than exists
today. The mechanism is unchanged, but the magnitude should is likely less on
current `main`.

rocketraman:
## 🚀 Description

Declare the Compose content subsurface's opaque region via
`wl_surface.set_opaque_region`, so the compositor can cull what is hidden behind
it.

The region is the window rect minus the four corner squares — the corners are
painted transparent by `applyFrameDecoration` so the drop shadow can show
through there, and claiming them opaque would clip it.

It is cleared whenever the resolved clear colour has alpha < 255, so genuinely
translucent windows (including `DecoratedWindow(transparent = true)`) are
unaffected, and it is only re-sent when the size, corner radius or tiling state
changes.

## 📄 Motivation and Context

Fixes #446 . Related to #444 (the Tao/Wayland resize artifact) — both concern how
the Tao backend's surfaces are presented during a resize, but they are
independent and this one stands alone.

Environment: Fedora 44, KWin 6.7.3 native Wayland, 3 × 4K at fractional scale
1.3, AMD Navi 48 / radv, Temurin 21.

Verified on the wire that the region is emitted with the corners excluded
(1024×720 window, KDE corner radius 5):

```
-> wl_compositor#36.create_region(new id wl_region#43)
-> wl_region#43.add(0, 5, 1024, 710)
-> wl_region#43.add(5, 0, 1014, 5)
-> wl_region#43.add(5, 715, 1014, 5)
-> wl_surface#35.set_opaque_region(wl_region#43)
```

Rendering is unaffected — rounded corners and drop shadow intact, no regressions
observed during resize, move, maximise or focus changes.

Measured during a resize drag on a 2842×1840 window:

| | before | after |
| --- | --- | --- |
| toplevel frame callback latency | 16.9 ms | 13.7 ms |
| median gap between window-edge moves | 25.6 ms | 20.0 ms |

**Caveat on those numbers:** they were taken before `de5e20ff` replaced the
`wl_shm` shadow subsurface with GTK's native CSD shadow, so they were measured
with one more full-window layer in the stack than exists today. The mechanism is
unchanged — the compositor still cannot cull beneath a surface it must assume is
translucent — but the magnitude should be re-measured on current `main` before
these figures are quoted anywhere.

## 📦 Types of changes

## ✅ Checklist

### DIFF
--- a/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
+++ b/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ffi/NativeTaoEglBridge.kt
@@ -121,6 +121,28 @@ internal object NativeTaoEglBridge {
         yLogical: Int,
     )
 
+    /**
+     * Wayland only: declares which part of the content surface is fully opaque,
+     * in surface (logical) units, with the four [cornerRadius]-sized corners
+     * excluded (they are painted transparent so the drop shadow shows through).
+     *
+     * Without this the compositor must treat our full-window surface as
+     * translucent and cannot cull the drop-shadow subsurface or GTK's toplevel
+     * underneath it — it alpha-blends all three every frame, which shows up as
+     * slower frame callbacks and therefore a slower resize. See
+     * `docs/linux-wayland-resize-latency.md`.
+     *
+     * Pass `logicalW <= 0` when the window really is translucent, which clears
+     * the region. Queued state — it lands with the next buffer commit.
+     */
+    @JvmStatic
+    external fun nativeSetOpaqueRegion(
+        handle: Long,
+        logicalW: Int,
+        logicalH: Int,
+        cornerRadius: Int,
+    )
+
     /** Pumps the back-buffer to screen via `eglSwapBuffers`. */
     @JvmStatic
     external fun nativePresent(handle: Long)
--- a/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/scene/TaoComposeSceneHostLinux.kt
+++ b/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/scene/TaoComposeSceneHostLinux.kt
@@ -222,6 +222,9 @@ internal class TaoComposeSceneHostLinux(
      */
     private var swapThread: SwapThread? = null
 
+    /** Last opaque region pushed: (logicalW, logicalH, cornerRadius). */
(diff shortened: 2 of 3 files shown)


kdroidFilter:
Great ! Thanks !

PR #456, opened 2026-08-05
fix(plugin): pathing jar for Linux jpackage classpath (JDK-8380085)

### LINKED ISSUE #454: jpackage Linux launcher SIGSEGVs at startup when the app has many jars (JDK ≤25)
## 🐛 Describe the bug

jpackage-generated Linux launchers segfault before the JVM starts once the app has enough jars on its classpath. The process dies with SIGSEGV inside `setenv()`, called from `jvmLauncherStartJvm` — no log output, no error dialog, nothing on stdout/stderr. The app simply never appears.

This reproduces with the **published Nucleus Demo v2.2.0** on Linux — confirmed on both the RPM and the AppImage — so it affects shipped artifacts, not just local builds.

Root cause is an upstream JDK bug, [JDK-8380085](https://bugs.openjdk.org/browse/JDK-8380085): `LinuxLauncher.c` forks a child that serializes a `JvmlLauncherData` blob and writes it to a pipe, and the parent reads it back with a single plain `read()` with no loop. A short read leaves the tail of the blob unfilled, so the `envVarNames` / `envVarValues` arrays hold uninitialized heap and `setenv()` dereferences garbage.

It is fixed in JDK mainline ([openjdk/jdk#30254](https://github.com/openjdk/jdk/pull/30254), commit `54c5e41`, 2026-03-20, adds `readFully`/`writeFully`) but **has not been backported to 25u**. Nucleus currently packages with JDK 25, so every Nucleus app on the jpackage path is exposed until that backport lands and we pick it up.

Filing here because Nucleus is the packaging plugin and can neutralize this for every app in one place — the expanded classpath is what pushes the payload over the line, and the plugin controls that.

## ⚠️ Current behavior

Launching the packaged app on Linux immediately produces:

```
Signal: 11 (SEGV)
Command Line: /opt/Nucleus Demo/bin/NucleusDemo

Stack trace:
#0  0x00007f607523e291 setenv (libc.so.6 + 0x20291)
#1  0x00005652e1a1e9b4 jvmLauncherStartJvm (NucleusDemo + 0x19b4)
#2  0x00005652e1a1e556 main (NucleusDemo + 0x1556)
#3  0x00007f6075221681 __libc_start_call_main (libc.so.6 + 0x3681)
```

`strace` catches the truncated transfer directly — the child writes 21885 bytes, the parent's single `read()` returns 5501 and it moves on:

```
child:  write(4, "8:\37\240\rV\0\0\17\0\0\0\2\0\0\0"..., 21885
parent: read(3,  "8:\37\240\rV\0\0\17\0\0\0\2\0\0\0"..., 21885) = 5501
parent: close(3)
        --- SIGCHLD {si_status=SIGPIPE} ---     # child killed writing the rest
        --- SIGSEGV {si_addr=0xfffffffffffcb600} ---
```

In the core dump, `envVarCount = 2` and the crash is on iteration 0, with both `setenv` arguments equal to the same uninitialized value:

```
rdi  0xfffffffffffe3680
rsi  0xfffffffffffe3680
rdx  0x1
```

`argv[0..2]` (exe, `-classpath`, the classpath itself) are intact; `argv[3..14]` — every `java-options` entry and the main class — all point at empty strings. The pointer offsets are all computed correctly; only the payload bytes are missing.

Two things worth noting, because they make this easy to misdiagnose:

- **There is no safe size threshold.** `read()` on a pipe returns as soon as *any* bytes are available, so this is a scheduling race. The `/opt` install here crashed with an expanded classpath of only **7674 chars across 90 jars**. A build can pass on one machine and crash on another with identical bytes.
- **The classpath itself is not corrupted.** It transfers fine. It's everything serialized *after* it that is lost, which is why the failure surfaces as a `setenv` crash rather than a classpath error.

## ✅ Expected behavior

Packaged Nucleus apps start normally on Linux regardless of how many jars are on the classpath.

Suggested fix: have the plugin's jpackage packaging collapse the `app.classpath=` entries in `<name>.cfg` into a single pathing jar whose manifest `Class-Path` lists its siblings, and point `app.classpath` at that one jar. This keeps the serialized payload small enough to transfer in one pipe write.

Verified on a copy of the installed v2.2.0 image: payload dropped **21885 → 1969 bytes** (now under `PIPE_BUF`) and the demo launched normally.

Notes on alternatives, both tested and rejected in a downstream app that hit this first:

- `@argfile` / `JDK_JAVA_OPTIONS` do not work — the launcher hands JLI a fully-assembled `argv`, so `@`-file expansion never runs. jpackage's own `@argfile` support applies to its build-time CLI, not the generated launcher.
- `app.classpath=$APPDIR/*` works but wildcard expansion order is unspecified. Images that ship the same artifact at two versions would resolve filesystem-dependently.

A pathing jar preserves entry order exactly and stays relocatable, since manifest entries resolve relative to the jar's own location — and it's the shape jpackage already expects (`--main-jar` plus a manifest referencing siblings).

The GraalVM native-image pipeline is unaffected, since it doesn't use the jpackage launcher at all.

## 💣 Steps to reproduce

1. On Linux, install the Nucleus Demo v2.2.0 RPM (the AppImage reproduces it too).
2. Run `/opt/Nucleus\ Demo/bin/NucleusDemo`.
3. It exits immediately with status 139 and no output. `coredumpctl info NucleusDemo` shows the `setenv` / `jvmLauncherStartJvm` trace above.

To confirm the mechanism rather than just the symptom:

```bash
strace -f -e trace=pipe2,read,write -s 40 /opt/Nucleus\ Demo/bin/NucleusDemo 2>&1 | grep -E 'write\(4,|read\(3,'
```

To confirm the fix, rewrite the installed image's `.cfg` to use a pathing jar and re-run — the app starts.

## 📱 Tech info

- OS: Fedora Linux (kernel 7.1.4), KDE Plasma / Wayland, x86-64
- Bundled runtime: JDK 25.0.3 (`lib/runtime/release`)
- Library/App version: Nucleus Demo 2.2.0 (rpm and AppImage)
- Classpath at failure: 90 jars, 7674 chars expanded
- Upstream: [JDK-8380085](https://bugs.openjdk.org/browse/JDK-8380085), fixed in mainline, not yet in 25u

kdroidFilter:
## Summary

- Work around [JDK-8380085](https://bugs.openjdk.org/browse/JDK-8380085): the Linux jpackage launcher can short-read the serialized classpath over a pipe and SIGSEGV in `setenv()` when many jars are listed in the `.cfg` (Nucleus Demo ~90 jars).
- After `jpackage` on Linux only, collapse multi-entry `app.classpath=` lines into a single pathing jar (`pathing-classpath.jar`) whose manifest `Class-Path` lists sibling jars in order.
- Matches the OpenJDK-documented workaround to “flatten the classpath”; macOS/Windows unchanged (bug is Linux-launcher-specific). GraalVM native images unaffected.

Fixes #454

## Test plan

- [x] Unit tests: `PathingJarClasspathTest` (collapse, no-op single entry, order, Linux layout, spaces encoding)
- [x] Manual: official nucleus-demo 2.2.0 AppImage rewritten with pathing jar starts; forced short-read no longer SIGSEGVs
- [ ] `createDistributable` / package Linux demo and launch the resulting app image
- [ ] Confirm `.cfg` has a single `app.classpath=$APPDIR/pathing-classpath.jar` and siblings remain on disk

### DIFF
--- a/plugin-build/plugin/src/main/kotlin/dev/nucleusframework/desktop/application/internal/PathingJarClasspath.kt
+++ b/plugin-build/plugin/src/main/kotlin/dev/nucleusframework/desktop/application/internal/PathingJarClasspath.kt
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2020-2026 JetBrains s.r.o. and respective authors and developers.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package dev.nucleusframework.desktop.application.internal
+
+import org.gradle.api.logging.Logger
+import java.io.File
+import java.util.jar.Attributes
+import java.util.jar.JarOutputStream
+import java.util.jar.Manifest
+
+/**
+ * Collapses multi-entry `app.classpath=` lines in jpackage `.cfg` files into a single
+ * pathing jar whose manifest Class-Path lists the sibling jars in order.
+ *
+ * jpackage's Linux launcher (JDK 25 and earlier) serializes the full expanded classpath into a
+ * JvmlLauncherData blob and transfers it parent-to-child over a pipe with a single plain
+ * read() and no loop. When the payload is large enough that the pipe returns a short read,
+ * envVarNames / envVarValues stay uninitialized and setenv() SIGSEGVs (JDK-8380085 /
+ * Nucleus issue 454). A pathing jar shrinks the launcher payload to a few KB (typically under
+ * PIPE_BUF), so the transfer always completes.
+ *
+ * Manifest Class-Path entries resolve relative to the pathing jar itself, so the layout stays
+ * relocatable. Order is preserved. Sibling jars remain on disk; only the `.cfg` classpath is
+ * rewritten.
+ *
+ * Safe to call when there is 0 or 1 classpath entry (no-op).
+ */
+internal object PathingJarClasspath {
+    /** Default name for the generated pathing jar; intentionally unlikely to collide. */
+    const val PATHING_JAR_NAME: String = "pathing-classpath.jar"
+
+    private const val CLASSPATH_PREFIX = "app.classpath="
+    private const val APPDIR_PREFIX = "\$APPDIR/"
+
+    /**
+     * Walks every launcher `.cfg` under [appJarDir] and collapses multi-entry classpaths.
+     *
+     * @return number of `.cfg` files rewritten
+     */
+    fun collapseInAppJarDir(
+        appJarDir: File,
+        logger: Logger,
+        pathingJarName: String = PATHING_JAR_NAME,
+    ): Int {
+        if (!appJarDir.isDirectory) return 0
+
+        val cfgFiles =
+            appJarDir
+                .listFiles()
+                .orEmpty()
+                .filter { it.isFile && it.extension.equals("cfg", ignoreCase = true) && it.name != "jvm.cfg" }
+                .sortedBy { it.name }
+
+        var rewritten = 0
+        for (cfg in cfgFiles) {
+            if (collapseCfg(cfg, appJarDir, pathingJarName, logger)) {
+                rewritten++
+            }
+        }
+        return rewritten
+    }
+
+    /**
+     * Locates the Linux jpackage `lib/app` directory under a createDistributable output root
+     * and collapses classpaths there.
+     *
+     * Layout: destinationDir/packageName/lib/app/ (launcher .cfg files)
+     */
+    fun collapseInLinuxAppImage(
+        destinationDir: File,
+        packageName: String,
+        logger: Logger,
+    ): Int {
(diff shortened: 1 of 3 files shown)


rocketraman:
Some comments:

PathingJarClasspath.kt:100 — when destinationDir has multiple children the collapse is skipped with no log line, so the crash would come back silently. A logger.warn there would be cheap.

PathingJarClasspath.kt:133 — entries are reduced to their basename via substringAfterLast('/'). jpackage emits flat $APPDIR/x.jar here (all 90 confirmed flat), but a nested entry would be silently rewritten to a path that doesn't resolve. Preserving the relative path would be strictly safer.

PathingJarClasspath.kt:192 — only spaces are percent-encoded; % or # in a jar name would corrupt the entry. Nucleus hash-mangles jar names, so unreachable in practice.

PR #499, opened 2026-08-10
fix(application): DecoratedWindow's outer-locals bridge shadows the scene's own LocalComposeSceneContext

mario:
## What's broken

Any `Popup`/`Dialog`/`DropdownMenu`/`Tooltip` composed inside a `DecoratedWindow` or `MaterialDecoratedWindow`'s content throws the moment it's shown, regardless of `nativePopupLayers`:

```
java.lang.IllegalStateException: LocalComposeSceneContext not provided
```

or, depending which code path gets there first:

```
java.lang.IllegalStateException: NavigationEventDispatcherOwner not found
```

Repro is trivial: put a Material3 `AlertDialog` behind a button inside any `DecoratedWindow`/`MaterialDecoratedWindow` content lambda and click it.

## Root cause

`TaoDecoratedWindowAdapter.Window` (`nucleus-application`) captures `currentCompositionLocalContext` from the parent `nucleusApplication { }` composition, which has no `ComposeScene` of its own since no Tao window has opened yet, and re-applies it inside the new window's own scene with a plain `CompositionLocalProvider(outerLocals) { content() }` wrapper.

That re-provides Compose's own internal `LocalComposeSceneContext` along with everything else, shadowing the real one this window's own scene already set up for itself. Every `Popup`-family composable resolves that shadowed local to pick which scene to render its layer into, and throws the instant it does. `nativePopupLayers` doesn't help, it only changes *which* scene type gets shadowed (`PlatformLayersComposeScene` vs `CanvasLayersComposeScene`), not whether the shadowing happens.

`TaoDecoratedDialogAdapter` already avoids this correctly: it bridges `outerLocals` through `DecoratedWindow`'s own `compositionLocalContext` parameter (a constructor-time scene property Compose applies *above* its own `LocalComposeSceneContext`, not below it, see `RootNodeOwner.setContent`) for the first composition, plus `LocalTaoCompositionLocalContextBridge` for every composition after. `TaoDecoratedWindowAdapter`, what `DecoratedWindow` and `MaterialDecoratedWindow` both actually route through, never got the same fix, even though `DecoratedWindow`'s own `compositionLocalContext` parameter doc comment already documents this exact use ("`DecoratedDialog` forwards its parent's locals here").

## The fix

Carries the dialog adapter's already-correct pattern over to the window adapter: `compositionLocalContext = outerLocals` on the initial `TaoDecoratedWindow(...)` call, plus `LocalTaoCompositionLocalContextBridge.current` in a `SideEffect` for every composition after. The existing Density/TaoWindow/TitleBarInfo/TextSelectionA11yPublisher snapshot-then-reprovide dance (there for an unrelated reason, see the file's own comment) is untouched.

## Testing

Added `TaoSceneOuterLocalsBridgeTest` at the `decorated-window-tao` scene-plumbing level: two real `CanvasLayersComposeScene`s standing in for the outer app-root composition and the window's own scene.

- `wrapping window content in outer locals with CompositionLocalProvider breaks Popup`, reproduces the exact crash (asserts `IllegalStateException`) using the old adapter's own shape.
- `bridging outer locals through the scene's own compositionLocalContext property does not break Popup`, confirms the fixed shape renders correctly.

Registered both in `TaoSceneTestBattery`/`TaoSceneTestBatteryDriftTest` per this module's own native-image coverage convention.

Full local run, both touched modules:

```
./gradlew :decorated-window-tao:test :nucleus-application:test
./gradlew :nucleus-application:detekt :decorated-window-tao:detekt :nucleus-application:ktlintCheck :decorated-window-tao:ktlintCheck :nucleus-application:apiCheck :decorated-window-tao:apiCheck
```

All green. No public API surface changed (`TaoDecoratedWindowAdapter` is `internal`), so `apiCheck` passes without a baseline update.

## How I found this

Chasing a real crash in a downstream Compose Multiplatform app built on Nucleus/Tao: `:app:run` showed no window at all at first (a separate, unrelated bug in that app's own launch ordering, an AWT Desktop init racing Tao for `NSApplication`, nothing to do with Nucleus), and once that was fixed, the very first `AlertDialog` opened from Settings took the whole process down with the stack trace above. Bisected to this adapter by comparing it against `TaoDecoratedDialogAdapter`, which handles the identical parent-locals bridging problem correctly.

### DIFF
--- a/decorated-window-tao/src/test/kotlin/dev/nucleusframework/window/tao/TaoSceneTestBattery.kt
+++ b/decorated-window-tao/src/test/kotlin/dev/nucleusframework/window/tao/TaoSceneTestBattery.kt
@@ -227,6 +227,12 @@ public object TaoSceneTestBattery {
             TaoSceneOuterLocalsBridgeTest()
                 .`bridging outer locals through the scene's own compositionLocalContext property does not break Popup`()
         }
+        run(
+            "TaoSceneOuterLocalsBridgeTest: bridged outer locals do not carry the outer layout direction into content",
+        ) {
+            TaoSceneOuterLocalsBridgeTest()
+                .`bridged outer locals do not carry the outer layout direction into content`()
+        }
         run("TaoSceneAnimationTest: tween advances exactly with virtual frames") {
             TaoSceneAnimationTest().`tween advances exactly with virtual frames`()
         }
--- a/decorated-window-tao/src/test/kotlin/dev/nucleusframework/window/tao/TaoSceneTestBattery.kt
+++ b/decorated-window-tao/src/test/kotlin/dev/nucleusframework/window/tao/TaoSceneTestBattery.kt
@@ -10,6 +10,7 @@ import dev.nucleusframework.window.tao.event.TaoSyntheticMouseWheelEventTest
 import dev.nucleusframework.window.tao.event.TaoWheelPinchZoomTest
 import dev.nucleusframework.window.tao.scene.TaoSceneAnimationTest
 import dev.nucleusframework.window.tao.scene.TaoSceneKeyboardTest
+import dev.nucleusframework.window.tao.scene.TaoSceneOuterLocalsBridgeTest
 import dev.nucleusframework.window.tao.scene.TaoScenePointerTest
 import dev.nucleusframework.window.tao.scene.TaoScenePopupTest
 import dev.nucleusframework.window.tao.scene.TaoSceneRenderTest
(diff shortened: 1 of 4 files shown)


mario:
The tao-tests (ubuntu-latest) failure is TaoSceneScrollTest > scroll direction is symmetric, unrelated to this change (nothing here touches scroll/pointer handling). Recent runs on main are all green, so this looks like a CI flake rather than something this branch introduced. Happy to help track it down if it turns out to be real, but wanted to flag it rather than leave it unexplained.

PR #596, opened 2026-08-27
fix(tao): forward macOS IME preedit into Compose

### LINKED ISSUE #595: Tao/macOS: IME marked text (preedit) never reaches Compose, and the commit Enter is delivered as a raw KeyDown — CJK input is unusable
## Summary

With the Tao backend on macOS, CJK input (tested with the Japanese IME) is currently unusable in Compose text fields, due to two related gaps:

1. **The composition (marked text / preedit) is never shown.** While composing, nothing appears in the text field — no inline underlined text, and no floating input window either.
2. **The Enter that commits a conversion is also delivered to Compose as a raw KeyDown.** In a multi-line `BasicTextField` this inserts a newline right before the committed text; in a single-line field it fires the IME action (e.g. triggers a search) mid-composition. Space / arrow keys / Backspace pressed during composition leak through the same way.

The same app works correctly on Compose Desktop / AWT.

## Environment

- Nucleus 2.5.0 (Tao backend, `nucleusApplication(backend = NucleusBackend.Tao)` + `DecoratedWindow`)
- Compose Multiplatform 1.12.0, Kotlin 2.4.10
- macOS (Apple Silicon), Japanese IME (Kotoeri, romaji input)

## Steps to reproduce

1. `nucleusApplication(backend = NucleusBackend.Tao) { DecoratedWindow { TitleBar {}; BasicTextField(...) } }`
2. Switch to the Japanese IME, focus the text field, type e.g. `nihongo`
3. → Nothing is displayed while composing (expected: inline preedit `にほんご` with underline)
4. Press Space to convert, then Enter to commit
5. → A newline is inserted, then the committed text `日本語` appears after it (expected: only the committed text)

## Analysis

I read the 2.5.0 sources and the bundled dylib, and I believe both symptoms are structural:

### 1. Preedit is never forwarded to the JVM

- The vendored tao `view.rs` implements `NSTextInputClient`, but `setMarkedText:selectedRange:replacementRange:` only updates the `marked_text` ivar — there is no `WindowEvent` variant and no JNI callback for preedit. The only IME-ish callback exported to the JVM is `onImeReplaceCommit`, which is dedicated to the press-and-hold accent picker.
- `TaoPlatformContext.startInputMethod` (`TaoComposeSceneHost.kt:1659-1692`) activates the input context and forwards the caret rect via `nativeSetImeRect`, but never calls `request.editText { setComposingText(...) }`. So `TextFieldValue.composition` stays `null` forever.
- Because the view claims `NSTextInputClient` conformance, AppKit's floating input window is suppressed as well — so the preedit is visible *nowhere*.
- For comparison, Compose Desktop's AWT path does this in `DesktopTextInputService2` (`inputMethodTextChanged` → `commitText` + `setComposingText`).

### 2. Raw keys are double-delivered during composition

- tao's macOS `key_down` queues `WindowEvent::KeyboardInput` **before** running `interpretKeyEvents:`, so even keys fully consumed by the IME (the committing Enter, conversion Space, candidate-navigation arrows, Backspace editing the preedit) are still delivered as raw key events.
- `TaoComposeSceneHost.onKeyEvent` (`TaoComposeSceneHost.kt:1222-1262`) forwards every `KEY_DOWN` unconditionally — there is no `hasMarkedText` / "consumed by IME" guard anywhere in the pipeline.
- AWT avoids this because `Component.dispatchEventImpl` routes key events through `InputContext.dispatchEvent` *before* key listeners, and consumed events are never delivered.

## Suggested direction

Roughly mirroring what `DesktopTextInputService2` does on AWT:

1. tao side: emit new JNI callbacks from `setMarkedText:` / `unmarkText` (e.g. `onImePreedit(text, selectionStart, selectionEnd)`), keeping `insertText:` → `ReceivedImeText` for commits.
2. tao side: while `hasMarkedText` is true (or when `interpretKeyEvents:` consumed the event), suppress the raw `KeyboardInput` — or flag it so the Kotlin side can drop it.
3. `TaoComposeSceneHost`: wire the new callbacks into the active `PlatformTextInputMethodRequest` via `editText { setComposingText(...) / commitText(...) / finishComposingText() }`.

## Notes

- This is a different layer from #558 (candidate-window positioning): the macOS caret-rect forwarding added there already works — the candidate window does appear at the caret. What's missing is the preedit round-trip into Compose and the key-event arbitration.
- App-side, we can partially work around symptom 2 by probing `hasMarkedText` via `objc_msgSend` from `onPreviewKeyEvent` and swallowing raw keys during composition, but symptom 1 cannot be worked around at all from the app, since no preedit data ever crosses the JNI boundary.

Thanks for the amazing project — happy to test a fix on a real Japanese-input setup.

kdroidFilter:
## Summary

- Fixes #595: CJK composition on Tao/macOS never reached Compose, and the committing Enter leaked as a raw KeyDown (newline before `日本語` in a multiline field).
- Native: `setMarkedText:` / `unmarkText` emit `ImePreedit`; `insertText:` during a preedit emits `ImeCommit`. Keys the IME consumed (and the matching keyUp) are not forwarded as `KeyboardInput`. Enter no longer cancels an in-flight composition.
- JVM: `TaoImeSession` routes preedit through `setComposingText` and commits through `commitText` (same shape as AWT `DesktopTextInputService2`). PressAndHold replace-commit stays on the same object.

## Test plan

- [x] `:decorated-window-tao:ktlintCheck` / `detekt` / `apiDump` (no public ABI change)
- [x] `:decorated-window-tao:test` including `TaoSceneImeTest` (offscreen protocol)
- [x] `:decorated-window-tao:taoHeadfulTest -Dnucleus.tao.headful.filter=#595` — real Kotoeri: `composition=TextRange(0, 3)`, commit `日本語`, no newline
- [ ] Re-run the headful Kotoeri case on a Mac with Japanese IME after review


takke:
> PS: I'd be glad to add a link to your application that uses Nucleus to my README :)

Following up on the README offer above — the migration has landed.

**ZonePane for Desktop v46.0.0** (released 2026-09-15) is now built on Nucleus 2.5.11 with the Tao backend, and
ships for **Windows, macOS and Linux**. The Desktop edition is currently distributed as a beta (download page:
https://zonepane.com/zonepane_beta.html), alongside our Android and iOS apps. Everything we worked through
together on the way — IME preedit on all three platforms (#596/#599/#603/#613), press-and-hold (#617), the
`WindowExceptionHandlerFactory` (#625) and the installer settings (#629/#630) — is in this release. Thank you
for the fast turnaround on every one of them.

We'd be glad to be listed under "Used by". The app itself is closed source, so a link to the website instead of
a GitHub repo would be great — please use the top page, since the beta download URL may move later:

- [ZonePane](https://zonepane.com/) — multi-SNS client (Bluesky / Mastodon / Misskey / Taittsuu) for Windows, macOS and Linux (beta), also on Android and iOS

kdroidFilter:
Thanks a lot! There, it's done!

Issue #576, opened 2026-08-19
Changing the window size trembling the whole compose content (TitleBar + Content) on both Win11 and Mac


janbarari:
## 🐛 Describe the bug
In one scenario I have to increase the window height when user moves between two different screens. with code below i was able to do that but the whole content trembling.
```
val windowHeightState = remember { mutableStateOf(SignInWindowHeight) }
    val animatedWindowHeight by animateDpAsState(windowHeightState.value)

    val windowState =
        rememberWindowState(
            size = DpSize(DefaultWindowWidth, SignInWindowHeight),
            position = WindowPosition.Aligned(Alignment.Center),
        )
```

https://github.com/user-attachments/assets/ab7a9813-dd1c-45c0-8b17-a93047e1d7e5

## 📱 Tech info
 - Device: AMD x64, Mac Os
 - OS: Win11, MacOs 13.7
 - Library/App version: 2.4.7

rocketraman:
Could be related to (or the same bug as) https://github.com/NucleusFramework/Nucleus/issues/444.

kdroidFilter:
thank you for your report, i really appreciate this attention to detail on your part, if i ever opened the #nucleus channel on the kotlin slack

rocketraman:
> Could be related to (or the same bug as) [#444](https://github.com/NucleusFramework/Nucleus/issues/444).

Nope, #444 is a different issue because it still happens with 2.5.1 which contains the fix for this issue (in fact, it seems even worse).

janbarari:
Thanks for the patch 🙌
I've tested, issue still presents.

**How to reproduce**:
Use decorated window with title bar, change the window height size with `animateDpAsState`.

PR #634, opened 2026-09-01
feat(tao): Compose window API v2 through an AWT-free clone

kdroidFilter:
## Summary

Compose 1.12's window API v2, for the Tao backend — through an **AWT-free clone of the package, `dev.nucleusframework.window.tao.v2`**, and only through it.

Compose's own `androidx.compose.ui.window.v2` is hard-wired to AWT: `Screen` wraps a `java.awt.GraphicsDevice` and reads insets via `Toolkit.getDefaultToolkit()`, and `WindowGeometryProviderScope` takes a displayable `java.awt.Window`. Neither exists on Tao, and reflection is off the table in a native-image runtime. Accepting those types would mean a surface where every scoped geometry provider — including the ones `requestSize` / `requestPosition` build internally — and `requestScreen` are silently dropped. That is worse than no surface, so **Compose's v2 types are deliberately not accepted anywhere** (an earlier revision of this PR did accept them best-effort; that layer is gone).

The clone mirrors the package member for member — `Screen`, `WindowScreenProvider(Scope)`, `WindowMetrics`, `WindowGeometryProviderScope`, the three providers with their companions, `WindowState`, `DialogState`, savers, `remember*` factories — backed by `TaoMonitors` + `TaoWindow`. Migrating from the Compose package is one import; deleting the clone restores the upstream import unchanged if JetBrains ever decouples its types. Every request is applied, `requestScreen` included, and `bounds` / `screenId` / `placement` / `isMinimized` are published back from the native window.

`DecoratedWindow` / `DecoratedDialog` / `HostedWindow` / `HostedDialog` and the `NucleusWindowHost` / `NucleusDialogHost` surfaces gain overloads for the cloned states (no default `state`, so v1 call sites stay unambiguous).

Supporting pieces:

- **`TaoMonitors` / `TaoMonitor`** — multi-monitor enumeration via a new `nativeGetMonitors` on each platform bridge (`EnumDisplayMonitors` + `GetDpiForMonitor`, `NSScreen.screens`, GDK), one tab-separated descriptor per monitor, physical px / top-left origin / work area, never empty. The AWT-free counterpart of `GraphicsEnvironment`.
- **Size and position stay split** (`CombinedBoundsProvider`) instead of folding into a `DpRect`: a rectangle cannot carry an unspecified position ("let the WM place it") or a wrap-content axis without turning both into `NaN`.
- **Wrap-content** providers (`Unconstrained`, `PreferredWidth/Height`) route through the window's own wrap-content path, which re-measures continuously, rather than a one-shot content measurement.
- **Observed geometry republishes on native move/resize** (conflated channel), not only when the v1 state changes — otherwise a WM-applied move left `bounds` stale for the window's life.
- **`NucleusWindowHost` / `NucleusDialogHost`** default bodies for both v2 flavours fall back to the v1 surface, so themed hosts keep their chrome; the default host overrides for the full path.

### Base-branch fixes folded in (each on `nucleus-2.6` first, then merged here)

- `nucleus-2.6` merged from `main` (#629 – #633).
- **`launcher-linux`: never autolaunch the session bus.** 9 of the last 9 `preMerge` jobs that hit the 30-min cap — on `main` too — were stuck in `:launcher-linux:test`: without `DBUS_SESSION_BUS_ADDRESS`, GDBus spawns `dbus-launch`, which waits for an X display that never comes. `get_connection` now treats "no address" as "no bus"; both native tests skip without one.
- `LcdTextTest` no longer asserts subpixel fringing on macOS (CoreText has none since Mojave) — the macOS `tao-tests` job had been red since #626.
- `#576` tremble case: an outer-only height skew is X server reporting lag, not tremble, unless the inner gate drifted too.

## Known limits

- `measureWindowContent` before the window exists (an *initial* provider, or a host that never exposes its window) has no scene to measure and falls back to the current content size clamped to the constraints. Once mapped it is a real `ComposeScene.measureContent` pass.
- X11 window managers apply their own placement policy to a client's *initial* position (openbox lands at 0,0); the centring headful case therefore asserts size and work-area containment there, and the exact centre on Windows/macOS.

Closed since the first revision: dialogs now feed `parentWindowMetrics` from their owner; an initial outer size is corrected by the measured insets once mapped; the v1 Absolute path waits for the X11 map before moving (the pre-map race that left a window at GTK's unallocated 1×1).

## Test plan

- [x] `NucleusWindowV2BridgeTest` (30) — provider evaluation (Fixed / Absolute / CenteredOnScreen / AlignedToScreen / scoped lambda / Unconstrained / PreferredWidth), `requestSize` keeps platform-default placement pre-window, screen provider fallback, savers, idempotent initial drain
- [x] `TaoMonitorsTest` — wire-format parsing, Dp conversion per scale, never-empty enumeration
- [x] `NucleusWindowHostTest` — clone `HostedWindow` forwards to the ambient host and reaches a v1-only host
- [x] **Headful e2e on a real window** (`taoHeadfulTest`, 11 cases, green on Windows / macOS / Linux CI, macOS re-run green): initial provider centring, `requestSize` / `requestPosition`, scoped provider reading live metrics and a real `measureWindowContent`, `requestScreen` landing on the target monitor, `screenId` tracking the hosting monitor — plus six edge cases: a burst of 40 position requests, size + position + screen in one tick, a bounds request on a maximized window, rapid maximize/restore toggling, requests from a background thread, a frame-paced move animation. The edge cases found and fixed three real bugs: `Current` read the not-yet-resized native window and a `requestSize` → `requestPosition` pair reverted the size (now resolved against the pending geometry); bounds applied during macOS's un-zoom animation were overwritten by its final frame (now apply-and-confirm); and the bridge's own `setMaximized(false)` — a `zoom:` toggle on macOS — re-zoomed the window or kept a toggle storm alive (the bridge never toggles zoom on macOS now; setting the frame un-zooms deterministically)
- [x] Full headful suite 32 run / 0 failed, three consecutive local runs; GraalVM native-image battery + headful green on all three platforms
- [x] Native built and exercised on Windows locally; Linux gdk3 code verified against the gdk 0.18.2 sources (`cargo fetch --target x86_64-unknown-linux-gnu`) after CI caught three blind-written mistakes
- [x] `check` (detekt, ktlint, apiCheck) on `decorated-window-tao` and `nucleus-application`; `apiDump` committed


decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ComposeWindowV2Bridge.kt
```diff
@@ -0,0 +1,327 @@
...
+        }
+        if (visible) {
+            ComposeWindowV2Access.setDialogInitialized(v2, true)
+        }
+    }
+}
+
+private fun currentScreenId(): String =
+    try {
+        GraphicsEnvironment
```
kdroidFilter: **[bug]** `currentScreenId()` (and `geometryPeerOrNull` at line 232) call `GraphicsEnvironment.getLocalGraphicsEnvironment()` on the Tao main thread. That is the XAWT init path this backend already documents as a deadlock: `TaoLinuxUriHandler` exists specifically because `java.awt.Desktop` / XAWT "initialises XAWT — its own X11 connection plus event-pump thread — and deadlocks against our GLX/Tao event loop", and `TaoScreenGeometry` exists to query the work area "without going through AWT's Toolkit/GraphicsEnvironment — neither exists on a no-AWT Tao runtime". `currentScreenId()` runs from `publishWindowObserved` for every v2 window once v1 reports an `Absolute` position (the usual post-map path), so this is not limited to custom providers. On Linux it can freeze the app; on GraalVM native-image it also pulls AWT into an otherwise AWT-free image.

**Suggestion:** Stop using AWT for screen id — `requestScreen` is already a no-op, so hardcode `"primary"` (the existing catch fallback) or use `TaoScreenGeometry`. Do not construct `java.awt.Window` / `GraphicsEnvironment` on the Tao thread. Resolve the common `Fixed` / `Absolute` / `Default` / `Current` providers without a dummy peer; if a custom lambda truly needs `WindowGeometryProviderScope`, evaluate it off-thread or against Tao geometry, not XAWT.

decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ComposeWindowV2Bridge.kt
```diff
@@ -0,0 +1,327 @@
...
+    val width = if (size.width.value <= 0f) Dp.Unspecified else size.width
+    val height = if (size.height.value <= 0f) Dp.Unspecified else size.height
+    return DpSize(width, height)
+}
+
+private fun evaluateBoundsProvider(provider: WindowBoundsProvider): DpRect? {
+    val dummy =
+        geometryPeerOrNull(
+            bounds =
+                Rectangle(
```
kdroidFilter: **[bug]** Non-`Default` providers are evaluated against a dummy peer at `(0, 0, 800, 600)`. Compose's `WindowBoundsProvider(sizeProvider = …)` / `requestSize` / `requestPosition` default the omitted axis to `WindowPositionProvider.Current` / `WindowSizeProvider.Current`, which read `windowMetrics.bounds`. Against this dummy that means position `(0, 0)` and size `800×600`. Compose's own AWT backend evaluates against the live window after it exists (`SwingWindow.initializeBounds` / `setBoundsFrom`), so `Current` means "keep the current geometry". Consequences: (1) `examples/tao-demo` switched from v1 `rememberWindowState(size = DpSize(1024.dp, 720.dp))` (platform-default position) to a size-only `WindowBoundsProvider`, which will open at the top-left instead of the WM default; (2) a later `requestSize(...)` also sends `positionProvider = Current` and will move the window to `(0, 0)`; (3) `requestPosition(...)` sends `sizeProvider = Current` and can reset size to 800×600. `BindWindowStateV2` reuses the same helper, so this is not only an initial-mapping issue. `ComposeWindowV2BridgeTest` only covers `WindowState()` (identity-equal to `Default`, short-circuited) and `WindowStateWithBounds` (Absolute+Fixed, which ignore dummy metrics).

**Suggestion:** Seed evaluation with the current v1/native geometry (and treat "no geometry yet" `Current` as `WindowPosition.PlatformDefault` / default size), matching Compose's live-window evaluation. Special-case size-only / position-only providers without going through a dummy. Add a test for `WindowBoundsProvider(sizeProvider = WindowSizeProvider.Fixed(…))` expecting `PlatformDefault` position, and one that `requestSize` does not clobber position.

decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/DecoratedDialogV2.kt
```diff
@@ -0,0 +1,80 @@
...
+        resizable = resizable,
+        enabled = enabled,
+        focusable = focusable,
+        onPreviewKeyEvent = onPreviewKeyEvent,
+        onKeyEvent = onKeyEvent,
+        compositionLocalContext = compositionLocalContext,
+        content = content,
+    )
+    BindDialogStateV2(state, v1, visible)
+    remember(v1, minSize, maxSize) { applyDialogSizeConstraints(v1, minSize, maxSize) }
```
kdroidFilter: **[bug]** Dialog `minSize` / `maxSize` are applied only by `remember { applyDialogSizeConstraints(...) }`, which mutates `v1.size` once after `DecoratedDialogV1` has already been composed (v1 always passes `minimumSize = null`). There is no `TaoWindow.setMinimumSize` / `setMaximumSize` call, unlike the window overload and unlike Compose's `SwingDialog` (`dialog.minimumSize` / `maximumSize`). Later `requestBounds` from `BindDialogStateV2` therefore ignore the constraints, and a `resizable = true` dialog can be dragged past them. The clamp also does not re-run when `v1.size` changes because `remember` keys only `v1` identity, `minSize`, and `maxSize`.

**Suggestion:** Pass min/max through to the underlying `TaoWindow` the same way `DecoratedWindowV2` does (`minimumSize` + `ApplyMaxSize` / `setMaximumSize`), and clamp incoming bound requests as well as the initial size. Apply the clamp before `DecoratedDialogV1` so the first map is already constrained.

decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ComposeWindowV2Bridge.kt
```diff
@@ -0,0 +1,327 @@
...
+    }
+}
+
+@Composable
+internal fun rememberWindowStateV1(state: WindowStateV2): WindowState = remember(state) { windowStateV2ToV1(state) }
+
+@Composable
+internal fun rememberDialogStateV1(state: DialogStateV2): DialogState = remember(state) { dialogStateV2ToV1(state) }
+
+internal fun minSizeOrNull(minSize: DpSize): DpSize? =
```
kdroidFilter: **[bug]** `minSizeOrNull` treats a partially specified `DpSize` (one axis `Unspecified`) as a real minimum. v1 `DecoratedWindow` then does `window.setMinimumSize(minimumSize.width.value.toDouble(), minimumSize.height.value.toDouble())`, so the unspecified axis becomes `NaN` on the JNI boundary. Native `SetMinInnerSize` only clears when `width < 0.0 || height < 0.0`; `NaN < 0` is false, so Tao gets a garbage constraint. Compose itself requires both axes (`DpSize.roundToDimensionOrNull` uses `isSpecified`), and `ApplyMaxSize` already requires both axes — this is inconsistent with both.

**Suggestion:** Return null unless both width and height are specified (same predicate as max size / Compose). If one-axis min is desirable later, translate the unspecified axis to `null`/`-1` in `setMinimumSize` and teach the native side to apply a single-axis constraint.

nucleus-application/src/main/kotlin/dev/nucleusframework/application/NucleusWindowHost.kt
```diff
@@ -71,6 +76,59 @@ public fun interface NucleusWindowHost {
...
+        nativeContextMenu: Boolean,
+        hiddenFromDock: Boolean,
+        minSize: DpSize,
+        maxSize: DpSize,
+        onPreviewKeyEvent: (KeyEvent) -> Boolean,
+        onKeyEvent: (KeyEvent) -> Boolean,
+        alwaysOnBottom: Boolean,
+        content: @Composable NucleusDecoratedWindowScope.() -> Unit,
+    ) {
+        DecoratedWindow(
```
kdroidFilter: **[suggestion]** The default v2 `Window` implementation calls `DecoratedWindow` directly instead of converting to v1 and delegating to the abstract `Window(state: WindowState, …)`. Existing themed hosts (the reason `NucleusWindowHost` exists) only implement the v1 SAM; a library `HostedWindow` with v2 state will silently skip their chrome. The KDoc says "themed hosts should override", but that is easy to miss and is a behaviour change for any host that has not been updated. Same pattern on `NucleusDialogHost.Dialog` v2. The new test overrides the v2 method, so it does not catch this.

**Suggestion:** Default implementation should `rememberWindowStateV1` + `BindWindowStateV2` and call `this.Window(… v1 …)` so existing hosts keep wrapping chrome. Document that `maxSize` is v2-only and dropped on that fallback, or add `maxSize` to the v1 host surface. Add a test whose host implements only the v1 method and assert v2 `HostedWindow` still hits it.

decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/DecoratedWindowV2.kt
```diff
@@ -0,0 +1,105 @@
...
+import androidx.compose.ui.unit.DpSize
+import androidx.compose.ui.unit.isSpecified
+import androidx.compose.ui.window.v2.WindowState
+import dev.nucleusframework.window.tao.DecoratedWindow as DecoratedWindowV1
+
+/**
+ * [DecoratedWindow] overload that accepts Compose Multiplatform 1.12's
+ * experimental window API v2 ([androidx.compose.ui.window.v2.WindowState]).
+ *
+ * Requested geometry (`requestBounds`, `requestPlacement`, …) is applied
```
kdroidFilter: **[suggestion]** Public KDoc says requested geometry (`requestBounds`, `requestPlacement`, …) is applied asynchronously, but `requestScreen` is drained and ignored. Callers using Compose's multi-monitor API will get a silent no-op. Draining is on a separate channel so it does not drop other requests (that part is correct); it just is not part of the public contract.

**Suggestion:** Document that `requestScreen` / `screenId` are not applied on Tao (primary work area only), on this overload and on the nucleus-application `DecoratedWindow` / `HostedWindow` v2 KDoc.

decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/ComposeWindowV2Bridge.kt
```diff
@@ -0,0 +1,327 @@
...
+
+private val defaultWindowSize = DpSize(800.dp, 600.dp)
+private val defaultDialogSize = DpSize(800.dp, 600.dp)
+
+internal data class ResolvedV2Bounds(
+    val position: WindowPosition,
+    val size: DpSize,
+)
+
+/**
```
kdroidFilter: **[suggestion]** The KDoc on `windowStateV2ToV1` restates the implementation and embeds design history (internal channels, `_bounds` / `_placement`, why Tao cannot live in `compose-ui`). Comments should explain a non-obvious constraint, not the architecture of the change.

**Suggestion:** Drop the second paragraph (the Java accessor file already documents the friend-package trick). Keep a one-liner that this snapshots pending v2 requests into the v1 state the existing window path consumes.

PR #664, opened 2026-09-08
feat(linux): after/before install/remove scripts

zaneschepke:
Add support for Linux install/uninstall scripts for JVM and GraalVM.

## 📄 Motivation and Context
There are many uses for this feature. My use case is to install and uninstall my app's daemon.

Tests are included in the PR. I also performed live testing with my packaged app and it was successful. 

## 📦 Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)

## ✅ Checklist


kdroidFilter:
Hi, Thanks for your work—it's a cool feature. Could you resolve the conflicts?

kdroidFilter:
Could you also tighten the KDoc on the four `linux { }` script properties? The pitfalls currently live in `CLAUDE.md`, which isn't what users (or the IDE) see.

Please document them on the properties themselves:

- **`afterInstall` / `afterRemove`** — concatenated after Nucleus's template, then run through electron-builder. `${sanitizedProductName}` / `${executable}` are substituted only when the token is single-quoted (`'${executable}-daemon.service'`). Double quotes (`"${executable}"`) are left unsubstituted and systemd hooks silently no-op.
- **`beforeInstall` / `beforeRemove`** — passed raw to fpm (`--before-install` / `--before-remove`). No electron-builder macro expansion at all; hardcode unit/path names (or don't use the macros).
- **Pacman** — `.INSTALL` `pre_remove` does not get deb-style `$1=upgrade`. Stop/disable the unit unconditionally and let `afterInstall` re-enable on upgrade.

Thanks.

kdroidFilter:
Two other bits to drop from this PR — they aren't the install/remove scripts:

**`jdk.jlink` (JEP 493) in `AbstractJLinkTask`**
That's a JVM runtime-image fix (Temurin 24+ can't emit an image that contains `jdk.jlink`). Real issue, but unrelated to Linux maintainer scripts. Please put it in its own PR.

**`nativeDistributions.appContent` → jpackage `--app-content`**
This looks like a way to ship a systemd `.service` next to the app. We already have `appResourcesRootDir` for extra files in the image; `afterInstall` can copy the unit into systemd from there. The new API is also incomplete: no KDoc, no tests, JVM-only, and a directory is flattened (subdirs ignored). Please drop it from this PR — if we still need `--app-content` later, that's a separate change.

This PR should stay: the four `linux { after/before Install/Remove }` properties, wired for JVM + GraalVM.

Issue #532, opened 2026-08-14
Window/Dialog fails to render when using Dp.Unspecified for wrap content height


janbarari:
## 🐛 Describe the bug
Regarding Kotlin Multiplatform using the `Dp.Unspecified` should let us have a window/dialog with wrap content height look but after i applied in Nucleus decorated window and dialog, it doesn't show anything
https://kotlinlang.org/docs/multiplatform/compose-desktop-top-level-windows-management.html#adapt-window-size-to-its-content

Code:
```Kotlin
DecoratedDialog(
    onCloseRequest = {},
    state = rememberDialogState(
        size = DpSize(300.dp, Dp.Unspecified)
    )
) {
    DialogTitleBar {
        Text("Doalog Title")
    }
    Text(
        text = "Dialog Content",
        modifier = Modifier.padding(40.dp)
    )
}
```

and this is the log after it shown
> 2026-08-14 17:14:12.657 java[5730:95227] CAMetalLayer ignoring invalid setDrawableSize width=600.000000 height=0.000000

## 📱 Tech info
 - Device: iMac
 - OS: 13.7
 - Library/App version: 2.4.4

Issue #569, opened 2026-08-17
nativePopupLayers: popups position against the window, not the screen (flip/clip uses a window-rooted work area)


kdroidFilter:
## Problem

`nativePopupLayers` works (the popup escapes the owner window, native placement is correct), but it is effectively unusable in real apps because **popup position decisions are made in window-rooted coordinates, not screen coordinates**.

The native side is fine: `nativeSetFrameInWindow` converts to screen via `ClientToScreen(parent)` + `boundsInWindow` (`nucleus_tao_windows_popup.c:730-734`, same model on macOS/Linux). The problem is upstream, in the geometry Compose uses to *decide* the position:

1. **`containerSize` is a size without an origin.** The layers report `containerSize = workAreaSize` (`TaoPopupSceneLayerWindows.kt:159-163`, `TaoPopupSceneLayer.kt:194-198`) so content can lay out larger than the owner window. But `Popup.skiko.kt` then flips/clips (`clipPosition`, on by default) inside `[0, workAreaSize]` **in window coordinates** — a virtual work-area-sized screen rooted at the window's client top-left.
2. **The work-area origin is discarded.** `TaoComposeSceneHostWindows.popupHost()` (`TaoComposeSceneHostWindows.kt:1463-1473`) reads only `w/h` from `nativeOwnerMonitorWorkArea` (`[x, y, w, h]`). No `TaoPopupHost*` API exposes the window's position on screen to the Kotlin side.

### Consequences

- Window near the bottom/right of the screen: a `DropdownMenu` near the real screen edge does **not** flip up (Compose believes there is `workAreaHeight` of room below the anchor) → the panel lands offscreen.
- A popup that should extend above/left of the window origin is clamped at `0` in window coordinates — the window edge acts as a fake screen edge.
- Positioning is only correct when the client origin coincides with the work-area origin (roughly: maximized on the primary monitor).

This is why the feature stays opt-in (`nativePopupLayers = false`) and unused outside demos.

## Proposed fix (pragmatic): screen clamp in `updateNativeFrame`

Clamp the final native frame into the real monitor work area at the single choke point `updateNativeFrame` (`TaoPopupSceneLayerWindows.kt:432`), converting to screen coordinates using the work-area origin that is already available natively but currently thrown away:

- Expose `workAreaOrigin` / parent-position-on-screen on `TaoPopupHost*` (all 3 OSes).
[...]

PR #674, opened 2026-09-17
fix(tao): centre wrap-content windows and dialogs once measured (#546)

### LINKED ISSUE #546: Window/Dialog not centered when size uses Dp.Unspecified
## 🐛 Describe the bug

Follow-up to #532 / #535. Wrap-content windows and dialogs now render when an axis is `Dp.Unspecified`, but `WindowPosition.Aligned(Alignment.Center)` is not applied after the real size is measured. The window/dialog does not appear at the center of the display.

Reported by @janbarari on https://github.com/NucleusFramework/Nucleus/pull/535#issuecomment-5309235758

```kotlin
val state = rememberWindowState(
    size = DpSize(400.dp, Dp.Unspecified),
    position = WindowPosition.Aligned(Alignment.Center),
)
```

#535 already re-centers **dialogs** on the parent once the wrap-content size is known. The same re-apply looks missing (or incomplete) for **windows** whose position is `WindowPosition.Aligned`, and possibly for dialogs when the alignment is relative to the display rather than the parent.

## Expected

After wrap-content measurement, the window/dialog should be placed according to `WindowPosition.Aligned(Alignment.Center)` using the resolved size.

## Actual

The window/dialog is shown, but not centered.

## Related

- #532 — original wrap-content blank-window bug
- #535 — fix that measures unspecified axes and applies `setInnerSize`

kdroidFilter:
## Summary

- Fixes #546: `DecoratedWindow` / `DecoratedDialog` with `Dp.Unspecified` on an axis rendered at the right size (#535) but were not placed by `WindowPosition.Aligned(Alignment.Center)` once that size was known.
- Root cause: the position effect resolved `Aligned` at first composition against the creation fallback (800×600, or NaN → 0 dp on Windows/Linux), and the `onMoved` echo then rewrote `state.position` to an `Absolute`, losing the alignment before wrap-content measurement arrived. The dialog re-centre from #535 was cancelled by the same echo and had no macOS branch.
- **Window**: the initial `Aligned` is remembered, skipped while wrap-content is unsettled, and re-applied after `setInnerSize` once the resize echo lands (macOS reads the live `NSWindow` frame).
- **Dialog with a parent**: one-shot re-centre on the parent when the measured size arrives, now on macOS too via `outerBoundsPx()`. The inner window receives `PlatformDefault` instead of the alignment so it does not also centre on the *screen*. On macOS the native pre-centre applies to the creation-size frame as well (no corner flash).
- **Parentless dialog**: centred on the screen, as AWT's `setLocationRelativeTo(null)`, through the window's own wrap-content path.
- Out of scope, from the issue comments: resizing the window when content changes later (Compose Desktop measures once) and `TitleBar` under an unspecified width.

## Test plan

- [x] `./gradlew :decorated-window-tao:taoHeadfulTest -Dnucleus.tao.headful.filter="#532|#546"` — 5 cases, 3 consecutive runs on macOS
- [x] With `src/main` stashed, `#546 window …` and `#546 dialog … parent` fail; `#546 parentless dialog …` passes on macOS only because tao centres new windows by default there
- [x] `:decorated-window-tao:test`, `ktlintCheck`, `detekt`
- [ ] Windows / Linux X11 headful run (no machine at hand)


kdroidFilter:
**CI triage** — the red checks are pre-existing on `nucleus-2.6`, none of them touch this change.

| Job | Failing cases on this PR | Same on the base branch |
|---|---|---|
| tao-headful (macos) | `#569` popup clamping ×3, `dialog appearance` (in-scene `Dialog` / M3 `AlertDialog`, native popup layer) ×3, satellite workspace stress, then the suite watchdog | [run 35206427078](https://github.com/NucleusFramework/Nucleus/actions/runs/35206427078) (2026-09-17): same `#569` ×3 + watchdog; [run 34045040435](https://github.com/NucleusFramework/Nucleus/actions/runs/34045040435) (2026-09-06): `#569` ×3 + `dialog appearance` + watchdog |
| tao-headful (windows) | `#569` ×2, `dialog appearance` / `graphicsLayer translation` native popup layer ×2 | [run 34045040435](https://github.com/NucleusFramework/Nucleus/actions/runs/34045040435): `#569` ×2 + native popup layer ×2 (4 failed) |
| test-graalvm (Linux / Windows / macOS) | headful under native-image: `#569` popup, satellite workspace stress, `satellite placement opened over a mapped parent` — 267 run, 3 failed | [run 35206427078](https://github.com/NucleusFramework/Nucleus/actions/runs/35206427078): the same 3 — 264 run, 3 failed |

The three extra cases in the counts are this PR's `#546` cases; they pass on every platform CI runs them on: macOS, Windows, Ubuntu X11 and the GraalVM Linux native image (`OK #546 …` ×3 in each log). The last green headful runs on the base are from 2026-09-04 ([run 33871865338](https://github.com/NucleusFramework/Nucleus/actions/runs/33871865338)), so the breakage landed between the 4th and the 6th.

PR #679, opened 2026-09-17
fix(tao): stop wake-served redraws from starving mouse input on Windows

kdroidFilter:
## Summary

Regression from #659 (`89c8cdea`): on Windows, tao-demo's Texture and Demo tabs keep painting but take no clicks.

- The Windows `UserEvent::Wake` handler has served the pending redraws itself since #659, so an embed's native modal loop keeps the window painting. A wake is a *posted* message, and Win32 hands posted messages out before hardware input. A frame served from the wake resumes the continuations that post the next wake, so an animating window (a `withFrameNanos` producer, an infinite transition) keeps the posted queue non-empty forever, and every `WM_MOUSEMOVE` / `WM_LBUTTONDOWN` starves behind it.
- The wake now serves frames only when `GetQueueStatus(QS_INPUT)` reports no input in the queue; otherwise the requests stay pending and the `MainEventsCleared` tick that follows the input serves them, as before #659. The paint-derived tick ranks below input, so that path never starved.

Bisected on Windows with synthetic clicks counted at the JNI event entry, natives rebuilt per commit:

| commit | clicks delivered |
|---|---|
| `main` `6b7a1092`, `410612a8`, `342b080d`, `5cd0553d`, `80d05639` | 3/3 on every run |
| `89c8cdea`, `a43f910c`, `8732c0df`, `e548f595`, `nucleus-2.6` tip | 0–1/3 |
| this fix | 3/3 on every run, redraw rate unchanged |

## Test plan

- [x] tao-demo Texture tab: 3 launches, 3/3 clicks each
- [x] `taoHeadfulTest -Dnucleus.tao.headful.filter="native view|#576|tab mouse click anywhere"` — 15 run, 0 failed, including the robot right-click context-menu cases the #659 wake change was written for
- [ ] macOS / Linux: `cfg(target_os = "windows")` only, no behaviour change there

Also cherry-picked onto #678 so that branch is testable; drop it there once this merges.


kdroidFilter:
Superseded by #678, which carried this exact commit as a cherry-pick (`2fda0695`) and is merged into `nucleus-2.6`. Nothing left to merge here.

PR #685, opened 2026-09-18
fix: pacman upgrade hooks

zaneschepke:
## 🚀 Description
Related to #664 pacman has separate hooks needed for pre/post upgrade.

This also fixed an issue on Linux where the payload install path is using display style names which keeps spaces. This breaks unquoted substitutions done by scripts and system unit templates. 

The fix is to use `packageName` over `appName` to prevent this on Linux.

## 📄 Motivation and Context
To complete pacman scripts support with the additional necessary hooks.

Manually install/upgrade tested.

## 📦 Types of changes

## ✅ Checklist


zaneschepke:
After more thought, that path change could potentially be breaking depending on the consuming application's setup for existing installs. I think it is the better approach, but might need some documentation or more consideration if that should be backed out. It seems 2.6 will have a lot of breaking changed so potentially not a super big deal if documented but worth considering.

kdroidFilter:
Thank you for reporting this; I will look into it.

Issue #654, opened 2026-09-06
Tao: distinguish trackpad panning from wheel scrolling in Compose


sargunv:
## ⚠️ Is your feature request related to a problem? Please describe

MapLibre Compose would like to bind trackpad panning to map movement and wheel scrolling to zoom. Under Nucleus 2.5.12, both arrive as ordinary Compose `Scroll` events, so the map cannot distinguish those gestures reliably. Fractional-delta heuristics are insufficient because trackpads can also emit whole-number deltas.

There is existing platform and Compose context for this distinction:

- Compose Multiplatform 1.12.0 defines [PanStart, PanMove, PanEnd, and panOffset](https://github.com/JetBrains/compose-multiplatform-core/blob/v1.12.0/compose/ui/ui/src/commonMain/kotlin/androidx/compose/ui/input/pointer/PointerEvent.kt). Its host-facing [ComposeScene.sendPointerEvent](https://github.com/JetBrains/compose-multiplatform-core/blob/v1.12.0/compose/ui/ui/src/skikoMain/kotlin/androidx/compose/ui/scene/ComposeScene.skiko.kt) accepts pan displacement. MapLibre Compose already handles these events.
- On macOS, [Tao's native handler](https://github.com/NucleusFramework/Nucleus/blob/v2.5.12/decorated-window-tao/src/main/native/vendor/tao/src/platform_impl/macos/view.rs#L1265-L1295) has precise-scroll information and native gesture phases. The current [Compose-facing path](https://github.com/NucleusFramework/Nucleus/blob/v2.5.12/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/TaoWindow.kt#L1224-L1235) exposes ordinary scroll deltas.

Related: #652 and #653 concern the sign and units of the existing scroll events; this request concerns distinguishing gestures. Versions investigated: Nucleus 2.5.12 and Compose Multiplatform 1.12.0.

## 💡 Describe the solution you'd like

Compose components should receive host-recognized trackpad panning through the standard Compose pan event API, separately from ordinary wheel scrolling. Applications could then configure those gestures independently without inferring device identity from scroll amounts.

## 🤚 Do you want to develop this feature yourself?

I can get to it perhaps next month if you don't by then; I'd be happy if you handled it though

Issue #653, opened 2026-09-06
macOS Tao: trackpad scroll magnitude includes display scale unlike AWT


sargunv:
## 🐛 Describe the bug

On macOS, Tao and AWT report different trackpad scroll magnitudes to the same Compose content. This is separate from the horizontal sign issue in #652.

## ⚠️ Current behavior

We observed different trackpad pan speeds while testing maplibre/maplibre-compose#1290. Source inspection shows a difference in how display scale enters the scroll conversion:

- [OpenJDK](https://github.com/openjdk/jdk/blob/jdk-25%2B36/src/java.desktop/macosx/native/libawt_lwawt/awt/AWTView.m#L391-L395) uses `scrollingDelta * 0.1`, without a display-scale multiplier.
- [Tao](https://github.com/NucleusFramework/Nucleus/blob/v2.5.12/decorated-window-tao/src/main/native/vendor/tao/src/platform_impl/macos/view.rs#L1265-L1272) converts logical AppKit deltas into physical pixels. [TaoWindow](https://github.com/NucleusFramework/Nucleus/blob/v2.5.12/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/TaoWindow.kt#L1224-L1235) then divides them by 10.
- The popup conversion in [MacOsWheelDelta.kt](https://github.com/NucleusFramework/Nucleus/blob/v2.5.12/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/event/MacOsWheelDelta.kt#L24-L31) also includes display scale.

For a 10-point AppKit delta on a 2× display, these conversions predict an absolute Compose scroll delta of 1 under AWT and 2 under Tao. They predict matching magnitudes on a 1× display. This ratio is derived from source, not a recorded comparison of identical physical gestures.

## ✅ Expected behavior

The same Compose scroll handling should produce the same movement under AWT and Tao, independent of display scale. AWT is the reference for the AWT-compatible scroll units Nucleus exposes.

## 💣 Steps to reproduce

1. Observe `PointerEventType.Scroll` and `change.scrollDelta` in identical AWT and Tao Compose content on a Retina display.
2. Compare vertical trackpad scrolling, which avoids the separate horizontal sign issue in #652.
3. Repeat on a 1× display to compare the effect of display scale.

## 📷 Screenshots

None; this concerns input deltas.

## 📱 Tech info

- Device: macOS desktop with a trackpad and Retina display.
- OS: macOS.
[...]

Issue #652, opened 2026-09-06
macOS Tao: horizontal trackpad scrolling is reversed relative to AWT


sargunv:
## 🐛 Describe the bug

On macOS with the Tao decorated window, two-finger trackpad panning is reversed on the horizontal axis. Vertical scrolling is correct. The same Compose content behaves correctly on AWT desktop and web.

The likely cause is a sign mismatch between two conversions:

- [Tao's macOS scroll handler](https://github.com/NucleusFramework/Nucleus/blob/v2.5.12/decorated-window-tao/src/main/native/vendor/tao/src/platform_impl/macos/view.rs#L1265) negates `scrollingDeltaX` but preserves `scrollingDeltaY`.
- [Nucleus's AWT conversion](https://github.com/NucleusFramework/Nucleus/blob/v2.5.12/decorated-window-tao/src/main/kotlin/dev/nucleusframework/window/tao/TaoWindow.kt#L1224) then negates both axes, assuming raw AppKit deltas.

## ⚠️ Current behavior

A two-finger swipe left on the trackpad produces a scroll delta with the opposite horizontal sign from what AWT reports for the same gesture, so content that pans with the scroll delta moves the wrong way. Vertical deltas match AWT.

## ✅ Expected behavior

Horizontal scroll deltas forwarded to Compose should have the same sign as under AWT, so a Compose app behaves identically on both hosts. Please account for Tao's horizontal sign conversion when forwarding macOS scroll events.

## 💣 Steps to reproduce

1. In a Compose app hosted in a Nucleus Tao window, observe `PointerEventType.Scroll` events (for example, log `change.scrollDelta`, or use any `Modifier.scrollable` with `Orientation.Horizontal`).
2. Two-finger swipe left on the trackpad.
3. Compare the sign of `scrollDelta.x` against the same gesture in an AWT-hosted window.

A concrete case: check out the branch from [maplibre/maplibre-compose#1290](https://github.com/maplibre/maplibre-compose/pull/1290) and run the demo with `mise run demo:desktop-nucleus`, then pan the map left/right with two fingers. The map moves in the opposite direction from the AWT demo (`mise run demo:desktop`).

## 📷 Screenshots

N/A

## 📱 Tech info

- Device: MacBook Pro (Apple M5 Pro)
- OS: macOS 26.6.2
- Library/App version: Nucleus 2.5.12, Compose Multiplatform 1.12.0

PR #692, opened 2026-09-22
fix(macos): make GraalVM native bundles signable and notarizable with Developer ID

takke:
## 🚀 Description

Developer ID signing of a GraalVM native-image `.app` fails in `packageGraalvm<Format>` on macOS, so a signed/notarized GraalVM DMG cannot be produced at all — and even once signing is made to work, Gatekeeper still rejects the notarized bundle. The JVM (jpackage) pipeline is unaffected. This PR fixes the three independent causes.

- **plugin `ExecutableType.kt`**: on macOS the executable-type marker is now written to `Contents/Resources/` (`findNativeBinaryDir` → `findMarkerDir`). Linux/Windows layout is unchanged (marker next to the binary).
- **runtime `ExecutableRuntime.kt`**: `readMarkerFile()` now looks for the marker next to the executable first (Linux/Windows, and macOS bundles produced by older plugin versions) and then in `../Resources/`. Implemented as `internal fun locateMarkerFile(executableDir)` so it is unit-testable; the public API is unchanged (`apiCheck` passes without a dump update).
- **plugin `AbstractElectronBuilderPackageTask.resignApp()`**: additionally signs every dylib / executable under `Contents/MacOS/` (recursively, main binary excluded — it is signed by the final bundle signature) with the Developer ID identity, mirroring what is already done for `Contents/runtime`. For JVM bundles `Contents/MacOS/` only contains the launcher, so this is a no-op there.
- **plugin `configureGraalvmApplication.kt`**: `fontconfig.bfc` is copied to `Contents/Resources/` instead of `Contents/MacOS/lib/` (`copyJawtToLib` now only ships `libjawt.dylib`; new `copyGraalvmFontConfig` task).
- **runtime `GraalVmInitializer.kt`**: on macOS, when `Contents/Resources/fontconfig.bfc` exists and `sun.awt.fontconfig` is unset, the property is pointed at it. `FontConfiguration` honours `sun.awt.fontconfig` before probing `<java.home>/lib`, so AWT font initialization keeps working; the old `<java.home>/lib/` layout still works unchanged.

## 📄 Motivation and Context

With `compose.desktop.mac.sign=true` + a notarization profile, `packageGraalvmDmg` aborts:

```
ZonePane.app: code object is not signed at all
In subcomponent: ZonePane.app/Contents/MacOS/.nucleus-executable-type
```

### Cause 1 — the executable-type marker is written into `Contents/MacOS/`

For GraalVM builds (no `.cfg` launcher) `updateExecutableTypeInAppImage()` writes the `.nucleus-executable-type` text marker next to the native binary, i.e. into `Contents/MacOS/`. `codesign` treats every file directly inside `Contents/MacOS/` as nested code, so a non-Mach-O file there makes bundle signing fail with the error above. (`codesign` tolerates non-code files in subdirectories such as `Contents/MacOS/lib/`, but Gatekeeper does not — see cause 3.)

### Cause 2 — `resignApp()` never re-signs the GraalVM dylibs

`resignApp()` only walks `Contents/runtime` (the jpackage JRE) and `Contents/Frameworks`. GraalVM bundles keep their dylibs (`libjvm.dylib`, `libawt*.dylib`, `libskiko-*.dylib`, …) in `Contents/MacOS/` and `Contents/MacOS/lib/`, so even with cause 1 fixed they would ship with the ad-hoc signature from `codesignDylibs`/`codesignBundle` and notarization would reject every one of them.

### Cause 3 — `Contents/MacOS/lib/fontconfig.bfc` makes Gatekeeper reject the notarized bundle

With causes 1 and 2 fixed the bundle signs, notarizes (Accepted, 0 issues) and staples, and `syspolicy_check distribution` / `gktool scan` are happy — but `spctl -a -t exec` still answers a bare `rejected` (`assessment:remote=true, verdict=false`). Bisecting by re-notarizing variants showed the only offender is `Contents/MacOS/lib/fontconfig.bfc`: Gatekeeper treats everything under `Contents/MacOS/` (subdirectories included) as code, and the notary service lists the file as a code object in the ticket too. `codesign` itself does not complain about it, which is why `--verify --deep --strict` passes.

The file lives there because `GraalVmInitializer` sets `java.home` to the executable directory and `sun.awt.FontConfiguration` reads `<java.home>/lib/fontconfig.bfc`.

Unit tests:

- New `ExecutableTypeMarkerTest` (plugin): macOS bundle → marker lands in `Contents/Resources/` and not in `Contents/MacOS/`; Linux/Windows layout → marker next to the binary; JVM `.cfg` present → no marker, `-Dnucleus.executable.type=` goes into the `.cfg` as before.
- `ExecutableRuntimeTest` (core-runtime): `locateMarkerFile` — next to the executable, `../Resources/` fallback, precedence when both exist, `null` when neither exists.
- `GraalVmInitializerTest` (graalvm-runtime): `resolveMacOsFontConfig` — found in `Contents/Resources/`, `null` when absent.
- `:core-runtime:check`, `:graalvm-runtime:check` and `plugin-build :plugin:{test,detekt,ktlintCheck}` pass for the touched code (the two pre-existing `plugin:test` failures on this machine — `LinuxSignerTest` needs gpg, `RenameMacAppBundleTest` hits APFS Unicode normalization — are unrelated).

Real-world check with a Compose Multiplatform app (ZonePane 46.1.0, arm64, macOS 26) using `compose.desktop.mac.sign=true` + `compose.desktop.mac.notarization.keychainProfile`, with plugin and runtimes published to mavenLocal from this branch:

- `notarizeGraalvmDmg` → BUILD SUCCESSFUL. Bundle layout: `Contents/MacOS/` and `Contents/MacOS/lib/` contain Mach-O files only; `.nucleus-executable-type` and `fontconfig.bfc` live in `Contents/Resources/`.
- All 16 Mach-O files under `Contents/MacOS/` (main binary, 13 dylibs, `lib/libjawt.dylib`, `lib/libskiko-macos-arm64.dylib`) carry `Authority=Developer ID Application …`, `flags=0x10000(runtime)` and a secure timestamp; `codesign --verify --deep --strict` passes.
- `notarytool`: **Accepted**, 0 issues; ticket stapled to the DMG (`stapler validate` OK).
- `spctl -a -t exec -vv ZonePane.app` → **`accepted` / `source=Notarized Developer ID`** (before cause 3 was fixed this was a bare `rejected` even though notarization had succeeded).
- `gktool scan` on a quarantined copy: "would be allowed but the user still needs to approve it on first launch" (the normal first-launch prompt for a notarized app).
- The native app launches and runs normally (no `Fontconfig head is null`, no font-related errors), i.e. `sun.awt.fontconfig` → `Contents/Resources/fontconfig.bfc` is picked up.
- JVM pipeline (`notarizeDmg`) with the same plugin: unchanged behaviour (same number of codesign invocations as before, Accepted, `spctl accepted`).

Follow-up (out of scope here): `copyAppResources` on macOS also places the app's `appResourcesRootDir` sidecar files into `Contents/MacOS/`. Any non-Mach-O sidecar will trip Gatekeeper the same way as `fontconfig.bfc` did (cause 3). The app used for verification ships no sidecars, so this PR does not touch it; it probably deserves the same treatment (move to `Contents/Resources/` and point `compose.application.resources.dir` there).

## 📦 Types of changes

## ✅ Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


takke:
CI note: `test-graalvm / GraalVM - macOS ARM64` is red, but the 4 failures are the Tao headful scroll-injection tests (#652 / #653 ×1 / #654 ×2, all `nativeDiagInjectScrollWheel returned false (window or content view gone?)`). The same 4 tests fail identically in the previous completed runs of this job on other branches (e.g. `feat/linux-install-scripts` on 2026-09-17, `feat/graalvm-headless-awt-skip` on 2026-09-08), and I could not find a green run of this job in the last ~80 workflow runs, so they look unrelated to this PR. `GraalVM - Windows x64` reports 0 failed tests and was cancelled by the macOS failure; `GraalVM - Linux x64` is green.

For what it's worth, the macOS job does exercise this change: `tao-native-test` is built with the new `copyGraalvmFontConfig` task and the native app starts and passes the remaining 31 headful tests.

kdroidFilter:
Great work, thank you very much!

PR #696, opened 2026-09-22
fix(tao): add cancellable system quit handling

anthonyhfm:
## 🚀 Description

Routes macOS Cmd+Q and Dock → Quit through the existing close-request path for every open Tao window. Each window receives its normal `onCloseRequest`, so applications can use the same save/confirm/cancel handling on macOS, Windows, and Linux. If no windows are open, the event loop exits.

This does not add a second application-level callback or public API.

## 📄 Motivation and Context

After migrating [Amethyst](https://github.com/anthonyhfm/Amethyst) to Tao, Cmd+Q bypassed the window close handlers and terminated immediately. That prevented applications from showing their existing save-before-close dialog.

- Added unit coverage for dispatching a close request to every open window without forcing an exit.
- Added unit coverage for exiting when no windows are open.
- Ran the affected Kotlin formatting, API, and unit-test tasks.
- Compiled the macOS Tao native libraries for Apple Silicon and Intel.
- Tested Amethyst through a local composite build on macOS: both Cmd+Q and Dock → Quit opened its existing unsaved-changes dialog through `onCloseRequest`.

## 📦 Types of changes

## ✅ Checklist


kdroidFilter:
thank you very much but could you make a pr to the nucleus 2.6 branch because I completely removed the awt backend in this branch and I quickly saw that your pr refers to it, sorry for not having indicated that it is the development branch

anthonyhfm:
@kdroidFilter yeah for sure. let me change that

anthonyhfm:
@kdroidFilter patch now only applies to TAO

kdroidFilter:
I’ve read through this and understand it’s an important feature, but I don’t see a Windows or Linux implementation—though it doesn't seem to be possible on Linux; that should be specified in the KDoc.

kdroidFilter:
Instead of a macOS-only onQuitRequest, could Cmd+Q / Dock → Quit just send a close request to every open window (like apple.eawt.quitStrategy=CLOSE_ALL_WINDOWS in AWT)? That way `onCloseRequest` works the same on all three platforms, with no second callback to know about.

anthonyhfm:
@kdroidFilter good suggestion. I've implemented it accordingly. Cmd+Q and Dock Quit now dispatch the existing close-request path to every open Tao window, so each window uses its normal onCloseRequest

kdroidFilter:
Maybe return `TerminateLater`, then reply true once `exitApplication()` runs, or false if windows are still open after the close requests. Also expose `TaoApplication.isQuitting` during that window so hide-to-tray onCloseRequest handlers can let a real quit through (same idea as Electron's before-quit).

kdroidFilter:
@anthonyhfm thanks, routing the quit through `onCloseRequest` is the right direction. I pushed a4801d8e on top of your commit to align it with Electron's macOS quit semantics — testing the first version end to end surfaced a few regressions.

**What the first version regressed**

| Case | 2.6 today | First version | Electron | Now |
|---|---|---|---|---|
| Logout / restart / shutdown (quit Apple event reply) | OK | `-128`: logout aborted at once, even when the app would quit | OK | OK |
| 3 windows that each remove themselves | quits | stays alive with no window | quits | quits |
| `TabWindows` + Cmd+Q | tabs kept | workspace emptied before exit (saved session is empty) | — | tabs kept |
| Unsaved doc window + `TabWindows` | quits | quits anyway (`onLastWindowClosed` bypasses the veto) | — | stays, tabs kept |
| Satellite palette | open | closed in the snapshot, even when the quit is cancelled | — | open |
| Main `onCloseRequest = ::exitApplication` + unsaved doc | quits | quits, doc lost | quit cancelled | quit cancelled |
| Close request closes its window and opens a "Save?" window | quits | dialog killed | waits for it | waits for it |

**Changes**

- **Native:** `-[TaoApp terminate:]` is overridden (Electron's `ElectronApplication`) instead of returning `TerminateCancel` from `applicationShouldTerminate`. The Apple event is answered OK, so loginwindow waits for the process instead of cancelling the logout. Once the event loop is gone, the real `terminate:` runs.
- **`TaoApplication.requestQuit`:** windows are asked newest first. The app exits once they all closed, and a window that stays open cancels the quit. A second request while one is in flight is ignored. A window opened during the quit defers the exit until it closes (`OnWindowAllClosed`).
- **`exitApplication()` from a close request during a quit** is that window's consent (like `app.quit()` while quitting), not an exit that overrides another window's veto.
- **Framework windows** (workspace satellites, tab windows, drag ghosts) are left out of the fan-out.
- **`TaoApplication.isQuitting` / `NucleusApplicationScope.isQuitting`** (Electron's `before-quit`): a hide-to-tray handler does `if (isQuitting) exitApplication() else hide()`. Without it, such an app no longer quits on Cmd+Q — same as Electron, documented in the KDoc.

**Verified** with a 59-case macOS e2e matrix, each case a real Tao app, compared against `nucleus-2.6` and your original commit:

- **Drivers:** Cmd+Q via `CGEventPostToPid`, Dock quit via `NSRunningApplication.terminate()`, the menu Quit item via AX, and logout / restart / shutdown quit Apple events with their replies.
- **Cases:** vetoes, async saves, dialogs, hidden / minimized / fullscreen windows, 50 windows, Cmd+Q bursts, a main-thread block during logout, a tab tear-off drag in flight, and a quit racing the close button.

Unit tests cover the state machine. `DockLandingRectTest` fails on `nucleus-2.6` too, unrelated.

PR #699, opened 2026-09-23
fix(linux): glibc 2.39 requirement crash

zaneschepke:
## 🚀 Description
Build natives using ubuntu-latest is risky as GitHub can bump this to a newer Ubuntu LTS version at any time.

The bigger issue: This makes tao require glibc version 2.39 on linux, which causes crashes on older, but still common linux distro versions (a few examples):
- Ubuntu 22.04 LTS (glibc 2.35)
- Devuan 5 (glibc 2.36)
- Debian 12 (glibc 2.36)

## 📄 Motivation and Context
App crashes on Devuan 5.

Using the latest Nucleus version and running my app on Devuan 5.

## 📦 Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

## ✅ Checklist
- [X] My code follows the code style of this project.


zaneschepke:
I should be clear on the testing bit:

I did not test this new pipeline runner change. I just tested the crash scenario.

kdroidFilter:
I understand the problem of depending on the glibc version of the runner, but using an older runner that will surely be deprecated soon, like 20.04, is not a solution either. Maybe we have to compile inside a docker container

kdroidFilter:
thanks a lot, I'll merge it anyway, when it's deprecated, I'll see what I do, I don't have the nerves to complicate the native build process now :)

